PR #50, opened 2022-02-08
Add a stiffer scalar advection version, convecting a Gaussian in a pe…

esclapez:
…riodic box.


esclapez:
Wrong description: this stiffer case advects a TanH profile of species or temperature.

PR #64, opened 2022-03-24
Check divu

nickwimer:
This branch implements the pressure rise due to enclosed mass injection by summing up velocity flux on all boundary faces and incorporates the correction in adjustPandDivU.


nickwimer:
Found a couple unused arguments... closing PR and updating...

PR #69, opened 2022-03-31
updated docs to include model description and enclosed domain ambient pressure formulation

nickwimer:
… pressure formulation


esclapez (approved):
Looks good, but it's difficult to read the eqs. I'll merge to make it easier.

PR #72, opened 2022-04-11
Landon/soot and spray

ldowen:
Adds functionality for soot and spray modeling when coupled with PeleMP, which is added as a submodule. Also adds time period based plot files, starting and regridding from grid files, and inputs for prevent plotting of reactions. Still need to figure out how to get derived soot variables to work.


Source/PeleLMAdvance.cpp
```diff
@@ -54,6 +60,11 @@ void PeleLM::Advance(int is_initIter) {
    std::unique_ptr<AdvanceAdvData> advData;
    advData.reset(new AdvanceAdvData(finest_level, grids, dmap, m_factory, m_incompressible,
                                     m_nGrowAdv, m_nGrowMAC));
+
+   for (int lev = 0; lev <= finest_level; lev++) {
+     m_extSource[lev]->define(grids[lev], dmap[lev], NVAR, amrex::max(m_nGrowAdv, m_nGrowMAC), MFInfo(), *m_factory[lev]);
+     m_extSource[lev]->setVal(0.);
```
esclapez, line 74: Do you need to setVal ?

Source/PeleLMAdvance.cpp
```diff
@@ -152,6 +173,22 @@ void PeleLM::Advance(int is_initIter) {
       averageDownnE(AmrNewTime);
 #endif
       fillPatchState(AmrNewTime);
+      // Reset external sources to zero
+      for (int lev = 0; lev <= finest_level; ++lev) {
+         m_extSource[lev]->setVal(0.);
```
esclapez, line 187: Same question here (is the setVal needed ?)

Source/PeleLMAdvance.cpp
```diff
@@ -54,6 +60,11 @@ void PeleLM::Advance(int is_initIter) {
    std::unique_ptr<AdvanceAdvData> advData;
    advData.reset(new AdvanceAdvData(finest_level, grids, dmap, m_factory, m_incompressible,
                                     m_nGrowAdv, m_nGrowMAC));
+
+   for (int lev = 0; lev <= finest_level; lev++) {
+     m_extSource[lev]->define(grids[lev], dmap[lev], NVAR, amrex::max(m_nGrowAdv, m_nGrowMAC), MFInfo(), *m_factory[lev]);
```
esclapez, line 73: Could m_extSource be smaller than NVAR ? Just trying to minimize how memory we use.

Source/PeleLMAdvection.cpp
```diff
@@ -741,12 +744,214 @@ void PeleLM::updateDensity(std::unique_ptr<AdvanceAdvData> &advData)
...
+
+void PeleLM::computePassiveAdvTerms(std::unique_ptr<AdvanceAdvData> &advData)
+{
+   //----------------------------------------------------------------
+   // Get the BCRecs and AdvectionTypes
+#ifdef SOOT_MODEL
+   int firstpass = FIRSTSOOT;
+   int numpass = NUMSOOTVAR;
+#else
+   int firstpass = 0;
```
esclapez: Not sure we need to define firstpass and numpass if we return right away.

Source/PeleLMEvolve.cpp
```diff
@@ -20,14 +20,21 @@ void PeleLM::Evolve() {
          amrex::Print() << "\n ====================   NEW TIME STEP   ==================== \n";
       }
 
+      bool regridded = false;
```
esclapez, line 23: Not sure what this bool is used for.

Source/PeleLMEvolve.cpp
```diff
@@ -80,8 +87,34 @@ PeleLM::writePlotNow()
 
    if ( m_plot_int > 0 && (m_nstep % m_plot_int == 0) ) {
       write_now = true;
+   } else if (m_plot_per > 0.0) {
```
esclapez, line 90: Thanks for doing that !
I was initially planning on pulling it out of incflo with the option of either doing it approximately at the m_plot_per or enforcing exactly the period by shortening the dt. We can probably update that later though ...

Source/PeleLMSetup.cpp
```diff
@@ -557,6 +605,75 @@ void PeleLM::variablesSetup() {
    } else {
       typical_values.resize(NVAR,-1.0);
    }
+   if (max_level > 0 && !m_initial_grid_file.empty()) {
```
esclapez, line 609: Could you wrap this piece of code in a function ?

Source/PeleLMAdvance.cpp
esclapez: And it seems to be defined regardless of whether we use SOOT/SPRAY or none. Is it voluntary ?

Utils/Make.PeleLMeX
```diff
@@ -25,6 +25,23 @@ ifeq ($(USE_EFIELD),TRUE)
    DEFINES += -DPELE_USE_EFIELD
    LMdirs += Source/Efield Source/Efield/GMRES Source/Efield/LinOps
 endif
+ifeq ($(USE_PARTICLES), TRUE)
+   ifdef SPRAY_FUEL_NUM
+      DEFINES += -DSPRAY_PELE_LM
```
esclapez: I'm guessing we can't easily change those to be more uniform with other defines (SPRAY_PELE_LM -> PELE_USE_SPRAY) ?

PR #79, opened 2022-05-17
Added spray derive vars

ldowen:
Added spray derived variables for plotting things like number of spray droplets per cell or the mass averaged diameter of the droplets in a cell.


esclapez (approved):
Looks great ! Thanks Landon.

PR #83, opened 2022-05-24
Add control over the number of digits in plt and chk file name

esclapez:
using `amr.file_stepDigits` (default is 5)


nickwimer (approved):
looks good

PR #84, opened 2022-05-24
Enable isothermal EB specification

esclapez:
At this point, activating isothermal EB with `peleLM.isothermal_EB = 1` set the entire EB as Dirichlet, the value of which then needs to be specified by the `setEBState` function, added to pelelm_prob.H.
See Exec/Regtests/EB_FlowPastCylinder for an implementation example.


esclapez:
- [x] enable both Dirichlet and homogeous Neuman (adiabatic wall) by adding control over the EB diffusion coefficient

Docs/source/manual/LMeXControls.rst
```diff
@@ -231,14 +264,14 @@ The `eb2.small_volfrac` controls volume fraction that are deemed too small and e
 This operation is done iteratively and the maximum number of iteration is prescribed by `eb2.maxiter`.
 For most applications, a single AMReX object is insufficient to represent the geometry. AMReX enable to combine
 objects using constructive solid geometry (CSG) in order to create complex geometry. It is up to the user to define
-the combination of basic elements leading to its desired geometry. To switch to a user-defined EB definition, one
+the combination of basic elements leading to his desired geometry. To switch to a user-defined EB definition, one
```
baperry2: Change "the user" to "users" and "his" to "their"

Source/PeleLMEB.cpp
```diff
@@ -384,6 +385,139 @@ void PeleLM::getEBDistance(int a_lev,
...
+               const amrex::Real* prob_lo = geomdata.ProbLo();
+               const amrex::Real xcell[AMREX_SPACEDIM] = {AMREX_D_DECL(prob_lo[0] + (i + 0.5) * dx[0],
+                                                                       prob_lo[1] + (j + 0.5) * dx[1],
+                                                                       prob_lo[2] + (k + 0.5) * dx[2])};
+               const amrex::Real xface[AMREX_SPACEDIM] = {AMREX_D_DECL(xcell[0] + ebfc_x(i,j,k) * dx[0],
+                                                                       xcell[1] + ebfc_y(i,j,k) * dx[1],
+                                                                       xcell[2] + ebfc_z(i,j,k) * dx[2])};
+               amrex::Real ebflagtype = 0.0;
+               setEBType(xface, ebflagtype, geomdata, *lprobparm);
+               ebdiff(i,j,k) = diff_cc(i,j,k) * ebflagtype;
```
baperry2, line 514: Just a note that if running with an LES model this will only get the molecular component of the thermal diffusivity because the turbulent component is added when moving from cc to fc. I believe this behavior is the same for how the no slip condition on EBs is presently implemented, and is fine. If using WALE and enough resolution at the boundaries the turbulent part should go to 0 anyway, and if not then we'd want to add a wall model.

baperry2 (approved):
Looks good, I tried one of the test cases and the results looked qualitatively reasonable

Source/PeleLMEB.cpp
esclapez: Good catch. It's probably worth adding a note in the doc for that.

PR #106, opened 2022-07-05
Add the TaylorGreen vortex breakdown case from PeleC

esclapez:
- add TG RegTests with processing scripts
 - initiate validation section of the doc with results from TG


Exec/RegTests/TaylorGreen/pelelm_prob.H
```diff
@@ -0,0 +1,115 @@
...
+
+    AMREX_D_TERM(const amrex::Real x = prob_lo[0] + (i+0.5)*dx[0];,
+                 const amrex::Real y = prob_lo[1] + (j+0.5)*dx[1];,
+                 const amrex::Real z = prob_lo[2] + (k+0.5)*dx[2];);
+
+    AMREX_D_TERM(const amrex::Real Lx = prob_hi[0] - prob_lo[0];,
+                 const amrex::Real Ly = prob_hi[1] - prob_lo[1];,
+                 const amrex::Real Lz = prob_hi[2] - prob_lo[2]);
+
+    AMREX_D_TERM(const amrex::Real x_c = prob_lo[0] + 0.5*Lx;,
```
ldowen: These don't seem to be used.

Exec/RegTests/TaylorGreen/pelelm_prob.H
```diff
@@ -0,0 +1,115 @@
...
+                 const amrex::Real y_c = prob_lo[1] + 0.5*Ly;,
+                 const amrex::Real z_c = prob_lo[2] + 0.5*Lz);
+
+    auto eos = pele::physics::PhysicsType::eos();
+    constexpr amrex::Real Pi = 3.14159265358979323846264338327950288;
```
ldowen: Would `M_PI` work instead?

Exec/RegTests/TaylorGreen/pelelm_prob.H
```diff
@@ -0,0 +1,115 @@
...
+
+    amrex::Real massfrac[NUM_SPECIES] = {0.0};
+    massfrac[O2_ID] = 0.233;
+    massfrac[N2_ID] = 0.767;
+
+    // TG functions
+    amrex::Real u[3] = {0.0};
+    u[0] = prob_parm.v0 * sin(prob_parm.omega_x * x / prob_parm.L) *
+           cos(prob_parm.omega_y * y / prob_parm.L) *
+           cos(prob_parm.omega_z * z / prob_parm.L);
```
ldowen: Only potential issue is that `z` won't be known if we are compiling in 2D. Not sure if that matters though.

esclapez: It's not a case intended for 2D, but it's worth making it possible to compile anyway, good point.

ldowen (approved):
Looks good!

PR #126, opened 2022-09-20
LES support and unity Le

baperry2:
This PR enables LES transport models for scalars and momentum in PeleLMeX. The turbulent viscosity is computed either using constant coefficient Smagorinsky or WALES. Turbulent transport for species and enthalpy are applied using this turbulent viscosity and constant assumed Schmidt and Prandtl numbers. 

A few notes on the implementation:
- Turbulent viscosity is computed based on the velocity field at the old timestep (N). Because the velocity field is not updated until the very end of the timestep, this means that when diffusion terms at N+1 are computed, the turbulent part fo the diffusion coefficient is still based on data for timestep N. This limitation would also be present for the IAMR implementation.
- Turbulent viscosity is computed at cell centers using centered differences to evaluate velocity gradients at the cell centers. This is so they can be directly added to the molecular diffusion coefficients, which are computed at cell centers. 
- EB compatibility not tested/supported yet

This PR also pulls in the unity Lewis number diffusion feature from LM, as that may often be desired for LES. 

Draft until documentation is added.

### DIFF
--- a/Source/PeleLM.H
+++ b/Source/PeleLM.H
@@ -145,8 +145,8 @@ class PeleLM : public amrex::AmrCore {
       LevelData (amrex::BoxArray const& ba,
                  amrex::DistributionMapping const& dm,
                  amrex::FabFactory<amrex::FArrayBox> const& factory,
-                 int a_incompressible, int a_has_divu, 
-                 int a_nAux, int a_nGrowState);
+                 int a_incompressible, int a_has_divu,
+                 int a_nAux, int a_nGrowState, int a_do_les);
 
       // cell-centered state multifabs
       amrex::MultiFab state;           // Stqte data: velocity, density, rhoYs, rhoH, Temp, RhoRT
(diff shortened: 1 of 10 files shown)


drummerdoc:
Should we worry about how you are computing gradients at coarse-fine boundaries?  If so, we could use the MLMG linear operators to supply those gradients consistent with how they are used in the diffusion solvers.

baperry2:
The velocity data gets FillPatched just before the turbulent viscosity is computed so I think what's there should be ok. 

But I might switch over to an MLMG operator - that would be safer and could also be better for handling EB stuff as well. Incflo uses an MLMG TensorOp to compute the velocity gradient tensor, but it computes turbulent viscosity directly at cell faces rather than cell centers. When I decided to switch over to computing coefficients at centers, this was the most expedient way to get the velocity gradient there.

baperry2:
Closing, superseded by https://github.com/AMReX-Combustion/PeleLMeX/pull/167

PR #136, opened 2022-11-04
Soret effect implementation

ThomasHowarth:
Implementation of mixture-averaged Soret effects for light species (W < 5), similar process to Wbar corrections for diffusion. Will also be a corresponding pull request in PelePhysics to obtain coefficients. Note there are some commented out personal diagnostics. @nickwimer @esclapez @drummerdoc @marchdf


esclapez:
Hi Thomas, thanks for doing this ! I'm catching up with some other blocking updates on the AMReX-Hydro side and when those are sorted out I'll give this a proper review and test.

esclapez (approved):
It looks good ! Thanks for doing this. I'll merge this now so that the testing machinery return to a running state, but it would be great if you could prepare a PR with some documentation of the implementation.

PR #137, opened 2022-11-04
Change wall BCs from reflect_even/odd to extrap/ext_dir.

cgilet:
@drummerdoc Note that velocity uses HOEXTRAP and scalars use FOEXTRAP. I just followed IAMR on that one.


cgilet:
Also, I realized I am not certain if PeleLM(eX) provides a default for ext_dir BCs. IAMR does so, but it is through NS, so PeleLM wouldn't inherit it.

esclapez:
Indeed, we're not setting a default zero ext_dir. I'll see how I can make that work.

esclapez:
Turns out, there is a default zero on ext_dir. This PR will introduce differences on the results of RegTest using walls.

esclapez (approved):

PR #141, opened 2022-11-18
Update spray test to use uniform spray initialize function, time was …

ldowen:
…also added as a spray variable


esclapez (approved):
LGTM

PR #167, opened 2023-01-10
LES Transport and Unity Lewis Numbers

baperry2:
This PR enables LES transport models for scalars and momentum in PeleLMeX. The turbulent viscosity is computed either using constant coefficient Smagorinsky or WALE. Turbulent transport for species and enthalpy are applied using this turbulent viscosity and constant assumed turbulent Schmidt and Prandtl numbers.

A few notes on the implementation:

- Turbulent viscosity is computed based on the velocity field at the old timestep (N). Because the velocity field is not updated until the very end of the timestep, this means that when diffusion terms at N+1 are computed, the turbulent part fo the diffusion coefficient is still based on data for timestep N. This limitation would also be present for the IAMR implementation.
- Turbulent viscosity is now computed at faces (based on velocity gradients from an AMReX TensorOp) and added to the face viscosity in the getDiffusivity function, just before applying zeroVisc function. 
- The TensorOp function doesn't support EB yet, so that capability will need to be added later
- This PR also pulls in the unity Lewis number diffusion feature from LM, as that may often be desired for LES.
- Documentation included, a suitable regtest will be added in a later PR
- This PR replaces https://github.com/AMReX-Combustion/PeleLMeX/pull/126

### DIFF
--- a/Source/PeleLMData.cpp
+++ b/Source/PeleLMData.cpp
@@ -6,7 +6,7 @@ PeleLM::LevelData::LevelData(amrex::BoxArray const& ba,
                              amrex::DistributionMapping const& dm,
                              amrex::FabFactory<FArrayBox> const& factory,
                              int a_incompressible, int a_has_divu,
-                             int a_nAux, int a_nGrowState)
+                             int a_nAux, int a_nGrowState, int a_do_les)
 {
    if (a_incompressible ) {
        state.define(  ba, dm, AMREX_SPACEDIM , a_nGrowState, MFInfo(), factory);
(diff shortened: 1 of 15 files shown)


Source/PeleLMTransportProp.cpp
```diff
@@ -1,12 +1,123 @@
...
+       amrex::ParallelFor(cp_cc, cp_cc.nGrowVect(), [=]
+                          AMREX_GPU_DEVICE (int box_no, int i, int j, int k) noexcept
+                          {
+                            getCpmixGivenRYT( i, j, k,
+                                              Array4<Real const>(state_arr[box_no], DENSITY),
+                                              Array4<Real const>(state_arr[box_no], FIRSTSPEC),
+                                              Array4<Real const>(state_arr[box_no], TEMP),
+                                              Array4<Real      >(cp_arr[box_no]) );
+                          });
+
```
esclapez, line 73: You'll need to add a Gpu::streamSynchronize() at this point because the multiarrays ParallelFor are non-blocking.

Source/PeleLMTransportProp.cpp
```diff
@@ -1,12 +1,123 @@
...
+         amrex::ParallelFor(ldata_p->visc_turb_fc[idim], ldata_p->visc_turb_fc[idim].nGrowVect(), [=]
+                            AMREX_GPU_DEVICE (int box_no, int i, int j, int k) noexcept
+                            {
+                              getTurbViscWALE( i, j, k, prefact,
+                                                Array4<Real const>(velgrad_arr[box_no]),
+                                                Array4<Real const>(dens_arr[box_no]),
+                                                Array4<Real      >(mut_arr[box_no]) );
+                            });
+       }
+
```
esclapez, line 111: Same Gpu::streamSynchronize() comment here.

Source/PeleLMData.cpp
```diff
@@ -17,6 +17,14 @@ PeleLM::LevelData::LevelData(amrex::BoxArray const& ba,
    press.define(   amrex::convert(ba,IntVect::TheNodeVector()),
                        dm, 1             , 1           , MFInfo(), factory);
    visc_cc.define( ba, dm, 1             , 1           , MFInfo(), factory);
+   if (a_do_les) {
+     for (int i; i < AMREX_SPACEDIM; ++i) {
```
esclapez: Need `int i{0};` otherwise I get uninitialized data.

esclapez (approved):
It looks good to me. Tested on a couple of simple cases successfully. 
Thanks Bruce !

PR #171, opened 2023-01-31
Update counterflow spray flame setup

esclapez:


baperry2 (approved):
This works for me now. Thanks!

PR #172, opened 2023-01-31
update GNUmake paths

baperry2:
Should enable compiling with the submodules obtained when using `git clone --recursive` without needing to edit the GNUmakefile or export any paths.


esclapez (approved):
LGTM, thanks Bruce !

PR #173, opened 2023-02-01
Add 2 new run-time diagnostics: PDF and conditional stats

esclapez:
@marchdf I'm thinking of moving this to PP and plugging this into PeleC at one point, so I'm interested in your feedback on the interface, maybe too convoluted ?
@baperry2 compared to what you're currently using, you'll need to manually specify a list of variables to output on the 2D plane instead of getting the entire state. There might be a way to add `macro` name that assemble set of variables of interest to make that easier.


marchdf:
Oh this is an excellent idea! Let me look at the code.

baperry2:
Lucas, this is fantastic! It's great to be able to dump derived variables using the diagnostics and its better to be able to specify what you need rather than just getting everything. The conditional statistics and PDF capability will also be quite useful.

One thing that would be very helpful is to add column headers to the output files and maybe use fixed width formatting to make them more easily human readable (not as important).

Additionally, while looking over this I've thought of small wish list of ways to extend/further improve this capability. All these are low priority, but are things I'd add as the use case comes up and I'm recording here so I remember:
- add option for fixed bounds for the conditioning variable similar to what there is for the PDF variable (to enable postprocessing combination of multiple time instances where the extrema are different)
- add condSq variable when computing integral conditional statistics (to allow computation of the standard deviation when combining data from multiple time instances during postprocessing)
- add option to internally accumulate conditional statistics/PDFs over time
- add problem-specific derived variables like in PeleC (https://github.com/AMReX-Combustion/PeleC/blob/d044eec3a6ad65b31b9088673701f37ff0e06a2c/Exec/RegTests/TGReact/prob.H#L151) to allow for custom use of these functions (the use case I have in mind is collecting statistics conditioned on radius for a turbulent jet.)

esclapez:
- [x] header in the ASCII files for PDF/Conditioned stats
- [x] fixed formatting of ASCII files
- [x] bounds on conditional variable
- [ ] accumulate stats over time
- [x] add user-defined derived

baperry2 (approved):
A couple of bugs fixed over here (https://github.com/AMReX-Combustion/PeleLMeX/pull/175) but once that's pulled in this is good to go

PR #175, opened 2023-02-10
Add diag fixes

baperry2:


esclapez (approved):
Thanks Bruce, this looks good. I wasn't sure initially about getting the avg. value of the conditioning variable in each bin, but I guess it is useful.

PR #178, opened 2023-02-21
Allow restart with fewer levels than previous checkpoint

baperry2:
The checkpoint file reader was failing if the desired number of levels in the restarted simulation is less than the number of levels in the checkpoint file, as the pressure and typical values get read in at the wrong location in the file. This PR fixes that.


esclapez (approved):
Good catch !

PR #190, opened 2023-03-27
Fix DiagFramePlane diagnostic

esclapez:


baperry2:
Nice. I ran into the issue with ghost cells on the Sydney flame case and worked around it by turning off interpolation. I'll test this fix on that case to make sure it works with interpolation now.

One more thing with diagnostics that I ran into is that if there is a derived variable with multiple components, you have to call out the individual component names rather than the derive name. I was going to look into relaxing that requirement.

esclapez:
I've been looking at this, and it might be tricky to get the diagnostics to interact with the derived because the intent was to have the diagnostics be as separate as possible in order to move them to PeleAnalysis at some point. I'll try something.

baperry2 (approved):
I verified that this fixes the issue I was having with interpolation for dumping planes.

baperry2:
> I've been looking at this, and it might be tricky to get the diagnostics to interact with the derived because the intent was to have the diagnostics be as separate as possible in order to move them to PeleAnalysis at some point. I'll try something.

It's not a big issue because one can simply list out all the components if desired. But it would be good to throw an error if someone requests a multi-component derive rather than the individual component, because right now that case is not handled properly.

Source/PeleLMDiagnostics.cpp
```diff
@@ -32,6 +32,15 @@ PeleLM::createDiagnostics()
         bool itexists = derive_lst.canDerive(v) || isStateVariable(v) || isReactVariable(v);
         if (!itexists) {
             Abort("Field "+v+" is not available");
+        } else {
+            if (derive_lst.canDerive(v)) {
+                const PeleLMDeriveRec* rec = derive_lst.get(v);
+                if (rec->numDerive() > 1) {
```
baperry2: This will capture when people request the individual derive components as well. The condition `rec->variableComp(v) < 0` should work to throw an error for the full multi-component derive but not the individual components.

esclapez: Indeed ! Good catch !

PR #193, opened 2023-04-06
Add a Tutorial on the FlameSheet case

esclapez:
- minor tweaks to the default input file of FlameSheet
 - minor tweaks to the stdout


Source/PeleLMPlot.cpp
```diff
@@ -53,7 +53,7 @@ void PeleLM::WritePlotFile() {
    const std::string& plotfilename = amrex::Concatenate(m_plot_file, m_nstep, m_ioDigits);
 
    if (m_verbose) {
-      amrex::Print() << " Dumping plotfile: " << plotfilename << "\n";
+      amrex::Print() << "\n Writting plotfile: " << plotfilename << "\n";
```
baperry2: Typo on writing

baperry2 (approved):

Source/PeleLMPlot.cpp
esclapez: Ah ah ! Indeed, not much cleaner than `Dumping` ...

PR #197, opened 2023-04-18
Add an `amr.regrid_on_restart` option.

esclapez:


baperry2 (approved):
Just curious - when would users want to set this option?

esclapez:
If you want to restart and add level(s) right away or change your tagging criterion right from the start. Otherwise, you will have to wait until the next `amr.regrid_int`. And even then, `regrid` is not recursive so that it can only adds a single level each time it is called even if you've increased your `amr.max_level` by 2.

PR #198, opened 2023-04-22
Add load balancing control

esclapez:
- add control over the PeleLMeX two dmap load balancing
 - add a brief overview of the code base
 - introduce a set of UserKeys to ease handling of string parameters (only used for load balancing as of this PR)


Docs/source/manual/Implementation.rst
```diff
@@ -0,0 +1,197 @@
...
+and Soret contributions. The later encapsulate the face-centered MAC velocities :math:`U_{ADV}`, the
+advection term :math:`A_{n+1/2,(k+1)}`, the pressure correction :math:`\chi` and a forcing container
+used in the RHS of advection/diffusion/reaction solves. In contrast with the `LevelData`, these two containers
+are freed at the end of the advance function, and are passed around in the functions called in `advance()`.
+
+Parallelism
+-----------
+
+*PeleLMeX* inherits the MPI+X approach from the AMReX library, where X can be any of CUDA, HIP or SYCL
+(Open-MP is currently unavailable) for heterogeneous architectures. The reader is referred to
```
baperry2: Rephrase, OpenMP is limited to nonreacting and explicit chemistry integration but nto supported for most reacting calculations.

Source/PeleLMRegrid.cpp
```diff
@@ -233,6 +481,67 @@ void PeleLM::ClearLevel(int lev) {
...
+        costMF.ParallelCopy(*mf,0,0,1);
+        for (MFIter mfi(costMF, false); mfi.isValid(); ++mfi)
+        {
+            a_costs[mfi] = costMF[mfi].sum<RunOn::Device>(mfi.validbox(),0) / mfi.validbox().numPts();
+        }
+    } else if (a_costMethod == LoadBalanceCost::UserDefinedDerivedSum) {
+        MultiFab costMF(a_costs.boxArray(), a_costs.DistributionMap(), 1, 0);
+        costMF.setVal(0.0);
+        std::unique_ptr<MultiFab> mf;
+        mf = derive("derUserDefined", m_cur_time, a_lev, 0);
```
baperry2, line 531: There's an edge case here where multiple variables are specified in "derUserDefined" and this requires that the one used for load balancing is the first. I guess any user that is using this functionality should be advanced enough to figure this out though.

baperry2 (approved):

Source/PeleLMRegrid.cpp
esclapez: Good point, I added a comment to the control section.

PR #216, opened 2023-05-31
Add spray sources to pltfile

esclapez:


Source/PeleLMPlot.cpp
```diff
@@ -210,6 +205,16 @@ void PeleLM::WritePlotFile() {
...
        plt_VarsName.push_back(spray_derive_name);
      }
+     if (SprayParticleContainer::plot_spray_src) {
+       plt_VarsName.push_back("spray_mass_src");
+       plt_VarsName.push_back("spray_energy_src");
+       AMREX_D_TERM(plt_VarsName.push_back("spray_momentumX_src");,
+                    plt_VarsName.push_back("spray_momentumY_src");,
+                    plt_VarsName.push_back("spray_momentumZ_src"));
+       for (const auto& spray_fuel_name : SprayParticleContainer::m_sprayDepNames) {
+         plt_VarsName.push_back(spray_fuel_name);
```
esclapez: @ldowen is this simply the name of the species ? If so we might want to prepend/append "spray_" and "_src" to distinguish it from other simple species name variables.

ldowen: Oh yeah. Good catch.

PR #233, opened 2023-08-21
Active Control max velocity default value fix

### LINKED ISSUE #232: Default value of active_control.velMax
Upon adding active control to my simulation case, the inlet speed kept going to zero resulting in a flash-back. 

I found out by looking at `PeleLM.H` that `m_ctrl_velMax` is set to zero by default, resulting in `Vnew` always being equal to zero:
```cpp
225     Vnew = std::max(Vnew, 0.0);
227     Vnew = std::min(Vnew, m_ctrl_velMax);
```

I don't have an issue anymore with using active control as I know I can just supply a value to make my case work, but perhaps something should be changed either in the TripleFlame example:
```bash
active_control.velMax = 1.0               # Optional: limit inlet velocity
```
Or in the code by adding a condition for line 227 of `PeleLMFlowControler.cpp` (slow ?) or setting an arbitrary large default value in `PeleLM.H` (not good practice ?). 

I would be appy to submit a pull request with the corresponding changes.

olivecha:
The default value for `m_ctrl_velMax` was set to -1.0 in `PeleLM.H`

The updated velocity in `PeleLMFlowControler.cpp` is not limited by the maximum velocity parameter if the parameter value is less than zero.

This result in the `active_control.velMax` input file parameter being truly optional (see issue #232).

Active control was able to converge without setting the `active_control.velMax` parameter:

image

### DIFF
--- a/Docs/source/manual/LMeXControls.rst
+++ b/Docs/source/manual/LMeXControls.rst
@@ -384,7 +384,7 @@ the user to set the inflow velocity. The following options are available when us
     active_control.height = 0.01              # [OPT, DEF=0.0] Where is the flame held ?
     active_control.v = 1                      # [OPT, DEF=0] verbose
     active_control.method = 1                 # [OPT, DEF=2] Controller: 1 - Linear, 2 - Quadratic, 3 - Weighted quadratic
-    active_control.velMax = 2.0               # [OPT, DEF=0.0] limit inlet velocity
+    active_control.velMax = 2.0               # [OPT, DEF=-1.0] limit inlet velocity, only used when positive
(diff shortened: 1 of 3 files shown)


baperry2:
Thanks for doing this!

Can you also modify the docs here to reflect the slightly updated behavior?
https://github.com/AMReX-Combustion/PeleLMeX/blob/62b496ca0abf7bad0ef971cb53fde6eb0226a781/Docs/source/manual/LMeXControls.rst?plain=1#L387C11-L387C11

Closes https://github.com/AMReX-Combustion/PeleLMeX/issues/232

baperry2 (approved):
Thanks again, this should be good assuming all the tests pass.

PR #235, opened 2023-08-29
Update PelePhysics - Use Sundials submod in gmake

baperry2:
This will fix an error where `make TPL` fails. Mirrors PeleC https://github.com/AMReX-Combustion/PeleC/pull/686


Utils/Make.PeleLMeX
```diff
@@ -1,7 +1,8 @@
-PELELMEX_HOME ?= /path/to/PeleLMeX
-AMREX_HOME ?= /path/to/AMREX
-PELE_PHYSICS_HOME ?= /path/to/PELE_PHYSICS_HOME
-AMREX_HYDRO_HOME ?= /path/to/AMREX_HYDRO_HOME
+PELELMEX_HOME ?= ../
+AMREX_HOME ?= $(abspath $(lastword $(PELELMEX_HOME)/Submodules/AMReX))
```
jrood-nrel: You will need a lowercase `amrex` path.

jrood-nrel:
I think the cloning in the CI files need to clone the submodules now as well

jrood-nrel:
We also need to make the CI more efficient here and have it use CMake soon.

baperry2:
Yeah it would be nice to have the CI match up better between C and LMeX

jrood-nrel:
Looks like `SUNDIALS_HOME` needs to be set in all the other CI files.

PR #240, opened 2023-09-07
Fix bug for isothermal EB on GPU

### LINKED ISSUE #237: peleLM.isothermal_EB = 1 causes CUDA error 700 - illegal memory access was encountered !!!
Hello, I am able to run a low Mach case with a fairly complex geometry with EB on CPUs. However, the same setup doesn't run on GPUs. It complains of illegal memory access. My case has isothermal walls.
After ruling out all possibilities, I have narrowed down the issue: switching the flag to peleLM.isothermal_EB = 0 runs seamlessly, but setting it to 1 results in the above mentioned error. 

The Backtrace file points to PeleLMUtils.cpp (line 333 - function - intFluxDivergenceLevelEB) where this issue may have arisen first. 

Is there a fix for this? Looking forward to some help on this.

Thanks!

Taaresh

baperry2:
Closes #237 

@taaresh1 can you verify that this solves your problem?

### DIFF
--- a/Source/PeleLMUtils.cpp
+++ b/Source/PeleLMUtils.cpp
@@ -334,8 +334,7 @@ void PeleLM::intFluxDivergenceLevelEB(int lev,
    geom[lev].GetVolume(volume);
 
    // Get area
-   const Real* dx = Geom(lev).CellSize();
-   const Real tmp_dx = dx[0];
+   const GpuArray<Real, AMREX_SPACEDIM> dx = Geom(lev).CellSizeArray();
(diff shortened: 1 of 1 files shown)


taaresh1:
@baperry2  This resolved the issue! Thanks a lot for the prompt response.

hsitaram (approved):
I would have preferred to use `const auto dx = Geom(lev).CellSizeArray();`  instead of const `Real* dx = Geom(lev).CellSize();` which is safe inside parallelfors

Thank you @baperry2 and @taaresh1 for fixing this issue

PR #245, opened 2023-09-13
Assimilate PeleLMeX to look more like PeleC

jrood-nrel:


jrood-nrel:
Still lots of stuff to iterate on, but this has the general framework in place, and I/we can slowly iterate on converting the current tests to CTest.

marchdf (approved):

PR #257, opened 2023-09-19
Renaming LM -> LMeX

marchdf:


jrood-nrel (approved):

baperry2:
This will require several updates in the Docs:

https://amrex-combustion.github.io/PeleLMeX/manual/html/Implementation.html#overview-of-source-code
https://amrex-combustion.github.io/PeleLMeX/manual/html/Tutorials_HotBubble.html#problem-specifications
+ similarly in other tutorials
and potentially other places...

marchdf:
yikes good catch. Wondering if this is a good idea... I did a second pass and I think I caught a bunch more. There's still the input files though... they use controls like `PeleLM.`. But maybe that's ok? Or we break everyone's input file.

marchdf:
Looks like I need https://github.com/AMReX-Combustion/PelePhysics/pull/422 first because https://github.com/AMReX-Combustion/PelePhysics/pull/417 broke some things.

baperry2 (approved):
One consideration would be that we don't want to add friction to remaining LM users transitioning from LM to LMeX now that LM is no longer being supported. But I don't think this adds much to that burden. But I agree, let's not change the parmparse prefix.

PR #279, opened 2023-10-04
Fix codeql warnings

marchdf:


Exec/RegTests/HITDecay/pelelmex_prob.cpp
```diff
@@ -126,9 +126,9 @@ PeleLM::readProbParm() // NOLINT(readability-make-member-function-const)
   ofs.close();
 
   // Read initial velocity field
-  const int nx = local_prob_parm.input_resolution;
-  const int ny = local_prob_parm.input_resolution;
-  const int nz = local_prob_parm.input_resolution;
+  const size_t nx = local_prob_parm.input_resolution;
```
jrood-nrel, line 129: I think this will show clang-tidy issues.

jrood-nrel (approved):

Exec/RegTests/HITDecay/pelelmex_prob.cpp
marchdf: I am actually curious about that. We do this in PeleC and it doesn't complain... But I feel like it should.

PR #283, opened 2023-10-13
fixes for Efield after renaming LM to LMeX

baperry2:
In https://github.com/AMReX-Combustion/PeleLMeX/pull/257 when file names were changed from LM to LMeX some of the Efield related file name changes were missed,  due to lack of test coverage for Efield capability. This PR gets the Efield cases working, other than `PremBunsen3DKuhl`, which relies on a mechanism that doesn't exist in PelePhysics.

TODO:

- [ ] Prior to the name change from LM to LMeX, a convention was mostly followed that files relating to the PeleLM class started with PeleLM, but files for other classes did not. Should we revert to that convention?
- [ ] Add test coverage for efield capability
- [x] Add guard because Efield + EB is not supported


marchdf (approved):

marchdf:
Well shoot maybe all that renaming was a bad idea?

baperry2:
Gonna merge this now and add the test coverage in a separate PR.

PR #286, opened 2023-10-17
Update docs to reflect current way of computing viscous stress

cgilet:
tensor.


drummerdoc (approved):
Perfect!  thanks Candace

PR #292, opened 2023-10-25
Update PeleLMeX_UserKeys.H

vtmaran:
Isotherm BC keyword


baperry2:
Thanks for submitting this @vtmaran. I believe that the user keys for isothermal boundaries were inadvertently changed in https://github.com/AMReX-Combustion/PeleLMeX/pull/204, and this brings them back in agreement with the documentation. 

Don't worry about the failing tests as they are unrelated. That's been an issue that has happened randomly lately and we are working on, but we will probably be able to merge this in the meantime.

baperry2 (approved):

PR #295, opened 2023-10-30
JOSS paper

esclapez:
In the process of wrapping up the JOSS revisions, let's bring this into the main branch before creating a new release.


baperry2 (approved):

baperry2 (approved):
Can we fix Marc HdF's name so his first and last names are split properly in the citation files, and curly bracket {Henry de Frahan} in the bibtex? 

Apparently with the `CITATION.cff` file, github will automatically generate a citation option  that generates bibtex in the right side menu, so we don't strictly need that in the README, but it doesn't hurt.

esclapez:
Indeed, it does add a APA/BibTex ref. link on the right, but this will make it more obvious :)

PR #313, opened 2023-11-16
Adding the radiation module

wjge:
Create the PR so that we can discuss it when we meet.
The build script:
https://github.com/AMReX-Combustion/PeleRad/blob/main/scripts/build_PeleLMeX_Frontier.sh


Source/PeleLMeX_Setup.cpp
```diff
@@ -14,6 +14,11 @@
 #ifdef PELELM_USE_SOOT
 #include "SootModel.H"
 #endif
+
+#ifdef PELELM_USE_RAD
+#include <PeleLMRad.hpp>
```
baperry2: I think this can be deleted as this header is not actually used here.

Source/PeleLMeX_Rad.cpp
```diff
@@ -0,0 +1,88 @@
...
+  BL_PROFILE_VAR_STOP(PLM_RAD_SOLV);
+
+  for (int lev = 0; lev <= finest_level; lev++) {
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (Gpu::notInLaunchRegion())
+#endif
+    for (amrex::MFIter mfi(*(m_extSource[lev]), amrex::TilingIfNotGPU());
+         mfi.isValid(); ++mfi) {
+      auto const& source_arr = m_extSource[lev]->array(mfi, RHOH);
+      rad_model->calcRadSource(mfi, source_arr, lev);
```
baperry2, line 82: I don't know whether this operation is expected to be computationally intensive, but it's weird that it is outside of any of the radiation profiling. Maybe extend PLM_RAD_SOLVE to also include this?

Source/PeleLMeX_Advance.cpp
```diff
@@ -120,6 +120,13 @@ PeleLM::Advance(int is_initIter)
     computeSootSource(AmrOldTime, m_dt);
   }
 #endif
+#ifdef PELELM_USE_RAD
+  if (do_rad_solve) {
+    BL_PROFILE_VAR("PeleLM::advance::rad", PLM_RAD);
+    computeRadSource(AmrOldTime, m_dt);
```
baperry2: Not necessarily for this PR: if I follow correctly, we are treating Radiation (and soot and spray) explicitly, 1st order, and decoupled from other physics. Should we recompute these source terms using the new state during each SDC iteration?

wjge: It depends on the time scale. The current implementation should work (for the radiation) for most combustion applications. However, for more extreme applications, e.g. Sandia has a case studying the nanosecond-scale spark ignition process, you are right we need to update it within the SDC iterations. I agree with you it would be great if we have that option, but it might not be necessary for this PR.

wjge:
There is a data file required in the test case:
https://github.com/AMReX-Combustion/PeleMP/blob/master/Exec/SootTests/PeleLMeX/laminar_flame/datafile_init/mueller_burner.dat

Please let me know where you want it to be.

baperry2:
That file isn't too big and we have similar pmf .dat files for the FlameSheet case. It can just go in the RegtTests/SootRadTest directory.

baperry2 (approved):

Exec/Make.PeleLMeX
```diff
@@ -54,6 +54,15 @@ ifeq ($(USE_SOOT), TRUE)
    DEFINES += -DNUM_SOOT_MOMENTS=$(NUM_SOOT_MOMENTS)
 endif
 
+ifeq ($(USE_PELERAD), TRUE)
+   $(info ******** Radiation model is activated ********)
```
marchdf: Not a huge fan of this kind of message. Can we remove the `*` here and below?

Exec/Make.PeleLMeX
```diff
@@ -54,6 +54,15 @@ ifeq ($(USE_SOOT), TRUE)
...
+   DEFINES += -DPELELM_USE_RAD
+   ifeq ($(USE_HIP), TRUE)
+       $(info ******** Radiation module is using hip ********)
```
marchdf: Do we need this define? PELERAD_USE_HIP == USE_HIP, so why cary an extra flag?

Exec/RegTests/SootRadTest/first-input
```diff
@@ -0,0 +1,112 @@
+#----------------------DOMAIN DEFINITION------------------------
```
marchdf: Can we unify the input file names? Something ending in `.inp` would be nice.

Exec/RegTests/SootRadTest/pelelmex_prob.H
```diff
@@ -0,0 +1,140 @@
+#ifndef PELELM_PROB_H
```
marchdf: Didn't we change this to `PELELMEX` ?

Exec/RegTests/SootRadTest/pelelmex_prob.H
```diff
@@ -0,0 +1,140 @@
...
+    massfrac[n] = amrex::max(0., amrex::min(1., pmf_vals[3 + n]));
+    norm += massfrac[n];
+  }
+  for (int n = 0; n < NUM_SPECIES; ++n) {
+    massfrac[n] = massfrac[n] / norm;
+  }
+  AMREX_D_TERM(state(i, j, k, VELX) = pmf_vals[0] * 1.E-2;
+               , state(i, j, k, VELY) = 0.;, state(i, j, k, VELZ) = 0.;);
+  amrex::Real rho_cgs, P_cgs;
+  P_cgs = prob_parm.P_mean * 10.;
```
marchdf: Avoid declaring and defining when you can do both: `const amrex::Real P_cgs = prob_parm.P_mean * 10.;`

Exec/RegTests/SootRadTest/pelelmex_prob.H
```diff
@@ -0,0 +1,140 @@
...
+  int j,
+  int k,
+  amrex::Array4<amrex::Real> const& beta,
+  amrex::GeometryData const& geomdata,
+  amrex::Box const& domainBox,
+  const int dir,
+  const int beta_comp,
+  const int nComp)
+{
+  amrex::ignore_unused(
```
marchdf, line 134: No need for this, just comment out the var names in the function arguments.

Exec/RegTests/SootRadTest/pelelmex_prob_parm.H
```diff
@@ -0,0 +1,16 @@
+#ifndef PELELM_PROB_PARM_H
```
marchdf: anotther use of PELELM

Source/PeleLMeX.H
```diff
@@ -32,6 +32,10 @@
 #include <AMReX_ErrorList.H>
 #include <AMReX_VisMF.H>
 
+#ifdef PELELM_USE_RAD
+#include <PeleLMRad.hpp>
```
marchdf, line 36: we have `hpp` files? hmmm we are going to have to clean that up at some point (later)

Source/PeleLMeX_Advance.cpp
```diff
@@ -120,6 +120,13 @@ PeleLM::Advance(int is_initIter)
     computeSootSource(AmrOldTime, m_dt);
   }
 #endif
+#ifdef PELELM_USE_RAD
```
marchdf, line 123: shouldn't this also be PELELMEX?

Source/PeleLMeX_Plot.cpp
```diff
@@ -116,6 +120,10 @@ PeleLM::WritePlotFile()
   ncomp += 1;
 #endif
 
+#ifdef PELELM_USE_RAD
+  ncomp += 3;
```
marchdf: also need to guard with `do_rad_solve` ?

Source/PeleLMeX_Rad.cpp
```diff
@@ -0,0 +1,88 @@
+#include <PeleLMeX.H>
+#include <PeleLMeX_Derive.H>
+
+#include <PeleLMRad.hpp>
+
+void
+PeleLM::RadInit()
+{
+  amrex::Print() << "Radiation model is activated "
+                 << "\n";
```
marchdf: just use `std::endl`

marchdf (changes requested):

Exec/Make.PeleLMeX
wjge: With the current clang version on Frontier, std::filesystem doesn't exist.
Instead, I have to do 

#ifdef PELERAD_USE_HIP
#include <experimental/filesystem>
#else
#include <filesystem>
#endif

The name may be confusing, this is considered a temporary fix to get cases running on Frontier. Once the clang compiler on Frontier moves std::experimental::filesystem to std::filesystem, the flag is no longer needed.

Source/PeleLMeX_Advance.cpp
wjge: This is following the style of all PELEMP macros: PELELM_USE_SPRAY and PELELM_USE_SOOT.

Source/PeleLMeX_Plot.cpp
wjge: There are two control flags for the radiation module, or any other PELEMP modules in general.
One is at compile time, and one is at runtime through the 'do_rad_solve' flag.

I considered 'ncomp +=3' belongs to the compile time, but it is fine if you think it fits to runtime.

Source/PeleLMeX_Rad.cpp
wjge: Notes
Use of std::endl in place of '\n', encouraged by some sources, may significantly degrade output performance.
from
https://en.cppreference.com/w/cpp/io/manip/endl

But if you insist, there is no problem to change it.

marchdf: Well that's a definite TIL for me! Thank you. It probably has no impact here because we don't actually do much IO but I can see where it might matter. For us, this would be a style decision. But I don't care that much

Source/PeleLMeX_Plot.cpp
marchdf: Ok I got that part, it's just that you seem to always use both the compile and the runtime guard. I was wondering if both were needed here as well to avoid other issues...

baperry2: I do believe that the runtime guard should be used here. Otherwise, if you compile but don't run with radiation, the multifab for plot variables will have 3 empty, unnamed components. 

It is pretty messy and bug-prone how we have 3 separate places in the code for counting pltvars, adding their names, and copying the data, and everything has to be in the same order for things to work. We should refactor this at some point in the future.

baperry2:
@marchdf @wjge are we ready to merge this?

Exec/RegTests/SootRadTest/pelelmex_prob.H
baperry2: Going to leave this here because this function uses ignore_unused in all the PeleLMeX cases

PR #314, opened 2023-11-16
Fix no EB production cases

### LINKED ISSUE #311: PremBunsen3D crashes
The defaults in these cases should be changed so they work out of the box.

### Discussed in https://github.com/AMReX-Combustion/PeleLMeX/discussions/309

<div type='discussions-op-text'>

<sup>Originally posted by **sjlienge** November 15, 2023</sup>
Hello,

The PremBunsen3D case crashes after printing the Build infos with SIGABRT. The same also applies for the PremBunsen2D. Is it known what the issue is? The source files seem alright to me. I would like to use the case to understand the setup of burner flames and setup a calculation of the Cambridge burner.

Thank you in advance already!

</div>

baperry2:
Don't compile with EB for Production cases that don't need it. Alternatively, we could still compile with EB but set `eb2.geom_type = all_regular` in the input files.

Closes #311

### DIFF
--- a/Exec/Production/CounterFlow/GNUmakefile
+++ b/Exec/Production/CounterFlow/GNUmakefile
@@ -2,7 +2,7 @@
 DIM = 2
 COMP = gnu
 PRECISION = DOUBLE
-USE_EB = TRUE
+USE_EB = FALSE
 USE_HYPRE = FALSE
 
 # Profiling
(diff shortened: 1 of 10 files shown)


jrood-nrel:
I would love to remove the EB ifdef and use all_regular like we do in PeleC sometime.

jrood-nrel (approved):

esclapez:
It won't work as well for LMeX (removing the EB ifdef). The linear solvers are completely different for EB, with different algorithms even when all_regular (especially the nodal solver). The plain linear solvers are faster than the EB versions and the EB versions are not fully available with kernel-fusing ParFor, so they tend to scale not as good. Probably nothing you couldn't fix with time though.

PR #326, opened 2023-12-13
Adapt to PeleRad in PelePhysics

jrood-nrel:


wjge:
GMRES didn't work for this chemistry previously.

Exec/RegTests/SootRadTest/kpDB/kpl_c2h4.dat
baperry2, line 1: Could delete these and point to the versions in PelePhysics instead. They aren't heavy and hopefully we will eventually make them not needed, so leaving them here is also ok.

jrood-nrel: I thought about that but a relative path that far away sounded like trouble and I didn't want to do anything fancy to try to account for it. If I have time I will just hard code them.

baperry2 (approved):

baperry2:
It's important that we keep the triple flame test because it is one of the tutorials. Let me check and see if I can figure out why it is failing.

PR #331, opened 2023-12-20
Update CMake for Soot, Spray, and Radiation

jrood-nrel:


CMake/BuildPelePhysicsLib.cmake
```diff
@@ -0,0 +1,239 @@
...
+                      ${PELE_PHYSICS_SOOT_DIR}/SootModel.cpp
+                      ${PELE_PHYSICS_SOOT_DIR}/SootModel_react.cpp
+                      ${PELE_PHYSICS_SOOT_DIR}/SootModel_derive.H
+                      ${PELE_PHYSICS_SOOT_DIR}/SootModel_derive.cpp
+                      ${PELE_PHYSICS_SOOT_DIR}/Constants_Soot.H
+                      ${PELE_PHYSICS_SOOT_DIR}/SootData.H
+                      ${PELE_PHYSICS_SOOT_DIR}/SootReactions.H)
+       target_include_directories(${pele_physics_lib_name} PUBLIC ${PELE_PHYSICS_SOOT_DIR})
+
+       if(PELE_PHYSICS_ENABLE_RADIATION)
```
baperry2: I think we should be able to build radiation without soot, so this shouldn't be nested within the soot conditional

Exec/RegTests/SootRadTest/pelelmex_prob.H
```diff
@@ -1,5 +1,5 @@
-#ifndef PELELMEX_PROB_H
-#define PELELMEX_PROB_H
+#ifndef PELE_PROB_H
```
baperry2: Why are these changed? They are `PELELM_PROB_H` for all other cases

Exec/RegTests/SootRadTest/pelelmex_prob_parm.H
```diff
@@ -1,5 +1,5 @@
-#ifndef PELELMEX_PROB_PARM_H
-#define PELELMEX_PROB_PARM_H
+#ifndef PELE_PROB_PARM_H
```
baperry2: Leave as `PELELM_PROB_PARM_H` to match other cases

Source/Efield/PeleLMeX_EFDeriveFunc.H
```diff
@@ -1,5 +1,5 @@
-#ifndef PELELMEF_DERIVEFUNC_H
-#define PELELMEF_DERIVEFUNC_H
+#ifndef PELEEF_DERIVEFUNC_H
```
baperry2: Another one that doesn't need to be changes?

README.md
```diff
@@ -108,18 +108,6 @@ If this is a clean install, you will need to make the third party libraries with
 
 Finally, make with: `make -j`, or if on macOS: `make -j COMP=llvm`. To clean the installation, use either `make clean` or `make realclean`. If running into compile errors after changing compile time options in PeleLMeX (e.g., the chemical mechanism), the first thing to try is to clean your build by running `make TPLrealclean && make realclean`, then try to rebuild the third party libraries and PeleLMeX with `make TPL && make -j`. See the [Tutorial](https://amrex-combustion.github.io/PeleLMeX/manual/html/Tutorials_HotBubble.html) for this case for instructions on how to compile with different options (for example, to compile without MPI support or to compile for GPUs) and how to run the code once compiled.
 
-When using CMake (cmake version >= 3.23 is required), first configure CMake directly in *PeleLMeX* root folder:
```
baperry2, line 111: These instructions are out of date, but I think we should have some instructions for cmake (i.e. just run the `cmake.sh` script)

Tests/CMakeLists.txt
```diff
@@ -209,19 +209,19 @@ endfunction(add_test_spray)
...
 #=============================================================================
 
-if(NOT PELELMEX_ENABLE_EB)
-  add_test_r(flamesheet-drm19-${PELELMEX_DIM}d FlameSheet)
-  add_test_rv(covo-${PELELMEX_DIM}d PeriodicCases)
-  add_test_r(hotbubble-${PELELMEX_DIM}d HotBubble)
-  add_test_rv(enclosedflame-${PELELMEX_DIM}d EnclosedFlame)
-  if(PELELMEX_ENABLE_RADIATION)
-    add_test_r(soot-rad-${PELELMEX_DIM}d SootRadTest)
+if(NOT PELE_ENABLE_EB)
```
baperry2, line 212: Can we run SootRadTest here? How about also including one (or more) of the EB test cases, perhaps EB_EnclosedFlame. If we compile, we might as well run these.

jrood-nrel: I think I will leave more test coverage to a separate PR. I think there is more to tackle with these CMake tests, but unfortunately I probably won't be able to do it.

baperry2: Yeah, that sounds reasonable.

baperry2 (approved):

PR #350, opened 2024-03-01
Incompressible fix/periodic channel/transport model consolidation

### LINKED ISSUE #337: Non-unity Lewis number
Hello,
Regarding the Lewis number. I would like to impose a constant number, for example, Le = 0.8.
I tried to consult the documentation, but I couldn't find an explanation on this matter.
Is it possible to set a non-unity Lewis number?

### LINKED ISSUE #340: pelelm.incompressible flag segfaults
Hi,
If I run with the incompressible flag switched on, the code segfaults. I tracked it down to the line in the Utils that tries to get the reactions MF without a flag check (e.g. m_do_react or m_incompressible, referenced below). I could fix it and put in a pull request but figured someone could add this quick fix without needing a full pull request.

https://github.com/AMReX-Combustion/PeleLMeX/blob/2b95bc8df51829d5f81a48bc7a9b060fc7fe8545/Source/PeleLMeX_Utils.cpp#L971

Thomas

ThomasHowarth:
This PR does the following:

1. Fixes the incompressible flag problem raised in #340 
2. Adds the ability to automatically adjust the pressure gradient for a periodic channel flow to maintain the mass flow rate given by the initial condition
3. Consolidates various transport models into a single function, including generalising the fixed diffusion coefficients, and adding in a scaling that is required for accurate Soret coefficients. This also addresses issue #337

### DIFF
--- a/Source/PeleLMeX_Setup.cpp
+++ b/Source/PeleLMeX_Setup.cpp
@@ -14,7 +14,6 @@
 #ifdef PELE_USE_SOOT
 #include "SootModel.H"
 #endif
-
 using namespace amrex;
 
 static Box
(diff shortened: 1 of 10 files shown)


Source/PeleLMeX_Utils.cpp
```diff
@@ -968,7 +968,10 @@ PeleLM::deriveComp(const std::string& a_name, Real a_time, int lev, int nGrow)
     // Get pressure: TODO no fillpatch for pressure just yet, simply get new
     // state
     auto* ldata_p = getLevelDataPtr(lev, AmrNewTime);
-    std::unique_ptr<MultiFab> reactmf = fillPatchReact(lev, a_time, nGrow);
+    std::unique_ptr<MultiFab> reactmf;
+    if (m_incompressible == 0) {
```
baperry2: This check should be on `m_do_react != 0`, otherwise multispecies nonreacting simulations will have the same segfault.

Source/PeleLMeX_Setup.cpp
```diff
@@ -330,26 +351,54 @@ PeleLM::readParameters()
 
   // -----------------------------------------
   // diffusion
+  ParmParse pptrans("transport");
+  pptrans.query("use_soret", m_use_soret);
   pp.query("use_wbar", m_use_wbar);
-  pp.query("unity_Le", m_unity_Le);
```
baperry2, line 334: Can you leave in a query for `unity_Le` and abort if it gets turned on with the message "unity_Le is depracated, used fixed_Le instead"? That would prevent anyone currently using the functionality from experiencing unexpected behavior after this gets merged.

baperry2 (changes requested):
Thanks for this @ThomasHowarth!

The code looks good! Just a few changes:

- [x] Handle the clang-tidy errors. We're a bit inconsistent about flags and have a mix of bools and ints. Bools in conditionals should just be like `if (boolvar)`, but ints should be like `if (intvar != 0)`. Sorry this is a bit annoying given the mix of different types.
- [x] Modify the documentation to account for the changes for constant Le diffusion. I think the only place this comes up is in the list of input file controls: https://github.com/AMReX-Combustion/PeleLMeX/blob/d30116e10b4d381329be4fa9bf041d5ec7f911c2/Docs/sphinx/manual/LMeXControls.rst?plain=1#L252
- [x] Also add the inputs for the periodic channel pressure gradient stuff to the controls section of the documentation
- [x] There's one test case that uses unity_Le and the input file should be updated to account for your changes. https://github.com/AMReX-Combustion/PeleLMeX/blob/d30116e10b4d381329be4fa9bf041d5ec7f911c2/Exec/RegTests/HITDecay/input.3d_les#L33

Source/PeleLMeX_Setup.cpp
ThomasHowarth: I have added this back in, and it will still run with unity Lewis number but will print a warning message also

ThomasHowarth:
I have made all the specified changes. It is perhaps a little convoluted when it comes to the various models possible. Currently, if we run without LES you can fixed Le or Pr. It doesn't need to be both, but both would obviously lead to fixed Sc. With LES, we _automatically_ run with fixed Pr/Sc (and hence fixed Le). Pr and Sc are specified in this case. Hopefully that makes sense, or happy to take suggestions.

baperry2 (approved):
Looks good, thanks again @ThomasHowarth

PR #366, opened 2024-04-12
Turb Inflow: move support codes to PelePhysics, add documentation

baperry2:
Partner PR of https://github.com/AMReX-Combustion/PelePhysics/pull/489

Removes turb file generating from the TurbInflow regtest and puts it in PelePhysics

Removes the python script from the HITDecay case and points to PelePhysics instead. Also removes old info about postprocessing from this README.

Adds documentation on using turbinflow


marchdf (approved):
Nothing quite like removing lines of code (even if they mostly moved somewhere else ;) )

PR #369, opened 2024-04-22
Boundary patch implementation

SreejithNREL:
Added functionalities for reporting species mass flux through user-specified patches on non-EB boundaries. Patches are user specific zones that the user can specify-currently, patch types are full-boundary, circle, rectangle, circle-annular and rectangle-annular. Once the variable do_patch_mfr is set to 1, the solver writes down these fluxes similar to other temporals.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Clone
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
     - name: Check formatting
       uses: DoozyX/clang-format-lint-action@v0.16.2
       with:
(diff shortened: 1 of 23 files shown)


baperry2:
A few general comments (I may have more when I dig in in more detail)

- [x] Add the new source files to `CMake/BuildPeleExe.cmake` 
- [x] Add some documentation on this capability . At a minimum, add the input file flags to `Docs/sphinx/manual/LMeXControls.rst`
- [x] Activate this capability for one of the test cases, e.g. in `Exec/RegTests/TurbInflow/input.3d`

baperry2 (changes requested):
Changes requested:

- [ ] Add documentation on additional input options
- [ ] In documentation, mention that bpatch fluxes are computed based on cell center locations on the coarsest level, so could differ from actual values when there is refinement
- [ ] Merge in fixes for host/device pointer stuff: https://github.com/SreejithNREL/PeleLMeX/pull/1

baperry2 (approved):

PR #391, opened 2024-06-21
Soret isothermal wall fix

### LINKED ISSUE #389: Wall boundary conditions (e.g. NoSlipWallIsothermal) do lead to wrong behaviour using the soret effect.
When the soret effect is turned on (simple, fuego), we can set up a test case, where species are "produced" at the wall due to zero gradient boundary condition. For soret, a zero gradient boundary conidition for the species is actually not correct, since the fluxes due to thermal diffusion have to be balanced by fluxes in the species profiles, leading to non zero gradient species profiles at the walls. I created a testcase (pipe with two walls on the top and the bottom with perodic bcs on the left and right) and set a temperature gradient from 300K to 700K from the bottom of the domain to the top (top wall temperature = 700K, bottom wall temperatur = 300K) and initialized the species profiles with a mixture of air and H2 (distributed equally over the domain). Over time the mass fraction of H2 increases and the mass fluxes of O2 and N2 decrease, which is physically not correct.
If you have any questions, I can also share the test case and we can discuss on this!

Thanks

ThomasHowarth:
This PR addresses #389 . Happy to discuss any of the implementation, and let me know if you can see spots for improvement/clarity. Currently, I've not implemented the necessary changes required for EB.


drummerdoc:
Is it possible to check in the test case?

Source/PeleLMeX_Diffusion.cpp
```diff
@@ -917,7 +993,7 @@ PeleLM::differentialDiffusionUpdate(
       GetVecOfConstPtrs(
         getDensityVect(AmrNewTime)), // this triggers proper scaling by density
       GetVecOfConstPtrs(getDiffusivityVect(AmrNewTime)), 0, bcRecSpec,
-      NUM_SPECIES - NUM_IONS, 0, m_dt);
+      NUM_SPECIES - NUM_IONS, 0, m_dt, GetVecOfConstPtrs(spec_boundary));
```
ThomasHowarth: This probably isn't correct for the boundary MF to be compatible with EFIELD, any suggestions?

ThomasHowarth: Having looked at the mechanisms with ions, the E species has a non-zero Soret coefficient (probably unsurprising), so this probably needs a little bit of thought

Exec/RegTests/IsothermalSoretChannel/pelelmex_prob.H
```diff
@@ -0,0 +1,143 @@
...
+AMREX_GPU_HOST_DEVICE
+AMREX_FORCE_INLINE
+void
+set_Y_from_Phi(ProbParm const& prob_parm, amrex::Real Y[])
+{
+  amrex::Real phi = prob_parm.phi;
+  auto eos = pele::physics::PhysicsType::eos();
+  amrex::Real Xt[NUM_SPECIES] = {0.0};
+  amrex::Real a = 0.0;
+  amrex::Print() << prob_parm.fuelID << std::endl;
```
baperry2: Remove print statements

Exec/RegTests/IsothermalSoretChannel/pelelmex_prob.cpp
```diff
@@ -0,0 +1,34 @@
...
+  pp.query("Thigh", PeleLM::prob_parm->Thigh);
+  pp.query("Tlow", PeleLM::prob_parm->Tlow);
+
+  PeleLM::prob_parm->bathID = N2_ID;
+  PeleLM::prob_parm->oxidID = O2_ID;
+  // get the fuel name from the peleLM.fuel_name declaration
+  amrex::ParmParse pp_pele("peleLM");
+  std::string fuelName = "";
+  pp_pele.get("fuel_name", fuelName);
+  amrex::Print() << "H2_ID 1 " << H2_ID << std::endl;
```
baperry2: remove print statement

baperry2 (commented):
@ThomasHowarth thanks for figuring this out and adding the test case. After this PR gets merged, I can submit a separate PR to have the case run in testing suite including a check on species balance.

I've been trying out the test case. I got rid of dt_shrink to take larger timesteps and make errors more visible. This fix does drastically reduce, but not quite eliminate, species imbalance errors tracked through `temporals/tempSpecies`. The remaining imbalance is small enough that I'm not very concerned, but I also don't get why conservation isn't quite exact (it doesn't seem to improve by tightening the diffusion solve tolerances).

A couple comments:
- Can the same fix be applied to the EB version of diffuse_scalar? That wouldn't eliminate the issue of isothermal EBs, but would at least allow EB simulations with isothermal domain boundaries and Soret.
- It appears that the updated BCs are not applied when diffusion fluxes are computed to compute divu. This should not affect species conservation and with the chi correction on divu, it should not affect the results in the limit of converged SDC iterations.

Source/PeleLMeX_Diffusion.cpp
baperry2: I'm open to the idea of just putting an abort for the combination of Efield+Soret+Isothermal and punting on this until someone needs that particular corner case.

drummerdoc: Agreed on the punt. :)

ThomasHowarth:
Sorry for the delay, busy month or so...

I've added in the changes so it is computed when computing divu, I've added in the same boundary input to the diffuse_scalar call when running with EB. Code also aborts on Isothermal+Soret+Efield. I am also still seeing some species flux in the species balance files, on the order of 3e-7 or so - any thoughts on this?

baperry2:
Sorry @ThomasHowarth, I know you've got a few big PRs in the queue waiting for review. I'm going to try to get to them this week.

I've dug into this one a bit and think I figured out the where the remaining conservation error is coming from. With the lagged correction approach, I was only thinking of the implicitly solved $`\tilde{D}^{n+1,k+1}`$ term in the update equation below, but, as you've now implemented the corrections must also be applied when computing $`D^n`$ and $`{D}^{n+1,k}`$, but those are explicitly computed and can't be lagged. Basically, the Soret and Wbar fluxes should be computed first in `computeDifferentialDiffusionFluxes` and those values should be used to set the inhomogeneous neumann conditions for the diffusive fluxes. For the Soret fluxes that should be easy with some code rearrangement. For the Wbar fluxes it's a little tricky because the Wbar fluxes at the isothermal boundary depend on the species gradients at the boundary, but also affect what the boundary condition needs to be for the species gradients, leading to an implicit definition of the boundary condition. I think it should be okay to compute the Soret terms first, use the boundary species gradients from the Soret terms in computing Wbar fluxes, then use the combined boundary gradients from both Soret and Wbar for the diffusive fluxes. Iterating around the Wbar flux computation would be the fully correct thing though.

Screenshot 2024-09-04 at 10 14 21 AM

Let me know if this makes sense. This is definitely a tricky problem and I'm not sure if I fully understand it. But I did hack a test for the Soret part of the fix (while turning Wbar fluxes off) and that did lead to correct mass conservation.

BTW, I also submitted #410 to make using temporals to track conservation a bit more user friendly.

ThomasHowarth:
Still working on this, and getting somewhere, but I noticed that in the addWbarTerm function that the flux [doesn't match with what you have above](https://github.com/AMReX-Combustion/PeleLMeX/blob/c8b1cab8564615258ad96e916d11f90330485104/Source/PeleLMeX_Diffusion.cpp#L464). I looks like 
$$\rho D_{k} Y_{k}\nabla \overline{W} / \overline{W}$$ 
rather than 
$$\rho D_{k} Y_{k}\nabla \overline{W} / W_{k}$$ 
which is what I would expect. Just want to double check this here.

Source/PeleLMeX_Diffusion.cpp
```diff
@@ -675,36 +784,14 @@ PeleLM::addSoretTerm(
 #pragma omp parallel if (Gpu::notInLaunchRegion())
 #endif
     {
-      FArrayBox rhoY_ed;
```
ThomasHowarth, line 675: There was an unnecessary rhoY_ed calculation in the Soret fluxes, which I've now removed

Source/PeleLMeX_Diffusion.cpp
```diff
@@ -604,17 +706,24 @@ PeleLM::addWbarTerm(
               for (int n = 0; n < NUM_SPECIES; n++) {
                 y[n] = rhoY(i, j, k, n) * rho_inv;
               }
-              amrex::Real WBAR = 0.0;
-              eos.Y2WBAR(y, WBAR);
-              WBAR *= 0.001;
+              amrex::Real imw[NUM_SPECIES] = {0.0};
+              eos.inv_molecular_weight(imw);
+              // amrex::Real WBAR = 0.0;
```
ThomasHowarth, line 711: I changed this based on the comment above

ThomasHowarth:
Based on what you said, I've made changes. The conservation error is now down at 10^-23 my run of the test case. I have also changed the wbar flux calculation based on the comment above. Additionally I removed an unnecessary rhoY edge calculation from the soret flux calculation.

Source/PeleLMeX_Diffusion.cpp
```diff
@@ -206,6 +206,171 @@ PeleLM::computeDifferentialDiffusionTerms(
...
+                                                 // proper wbar flux
+    Vector<Array<MultiFab, AMREX_SPACEDIM>> wbarfluxes(finest_level + 1);
+    for (int lev = 0; lev <= finest_level; lev++) {
+      for (int idim = 0; idim < AMREX_SPACEDIM; idim++) {
+        wbarfluxes[lev][idim].define(
+          grids[lev], dmap[lev], NUM_SPECIES, 1, MFInfo(), Factory(lev));
+        wbarfluxes[lev][idim].setVal(0.0);
+      }
+    }
+    int iter_max = 10;
```
ThomasHowarth: Currently 10 iterations hardcoded - we could do some sort of residual check instead

drummerdoc: 10 seems a bit excessive, no?

ThomasHowarth: Probably - based on the test case it converged in 2 or 3 iterations. EDIT: things are blowing up, will keep testing

ThomasHowarth:
Without the Wbar correction, I have also managed to eliminate the conservation error (Up to 10^-23). However, iterating around the wbar fluxes seems to blow up the gradients, and I'm not entirely sure why. I'll keep working on it, but I've also pushed the latest version so you can take a look if you like.

baperry2:
Nice progress a @ThomasHowarth. This has turned out to be much more involved than expected! 

If a solution doesn't jump right out at you for the exploding wbar iterations, do you think we can merge this without the wbar iterations (or maybe hardcoding to 1 "iteration") and then have a separate PR for the final correction? There's a lot here that I don't want to risk falling out of date and even without the last bit it's more correct than what's in there now.

Source/PeleLMeX_K.H
```diff
@@ -440,6 +440,31 @@ getMwmixGivenRY(
...
+  amrex::Array4<amrex::Real> const& gradMwmix) noexcept
+{
+  using namespace amrex::literals;
+
+  auto eos = pele::physics::PhysicsType::eos();
+  amrex::Real imw[NUM_SPECIES] = {0.0_rt};
+  eos.inv_molecular_weight(imw);
+  gradMwmix(i, j, k) = 0.0;
+  for (int n = 0; n < NUM_SPECIES; n++) {
+    imw[n] *= 1000.0; // CGS -> MKS
```
ThomasHowarth, line 585: I thought this was correct, but if I comment out, everything converges nice and quickly - thoughts?

drummerdoc: This reflects me being away from things for a while, but the MKS version of mw *could* be kg/kmole (depending on how it is used elsewhere) so that there is no conversion factor necessary.  :)

baperry2: That is weird because I do think multiplying by 1000 is correct here. When computing MWmix we divide by 1000. Perhaps there's some error cancellation going on?

ThomasHowarth:
I think perhaps the best course is to just do the Soret correction for the conservation, disable the use_wbar for consistency (in these cases) and leave a warning/todo note. If this seems reasonable I can tidy things up and leave it like this for now? As you say, at least it will work, if not perfect yet.

ThomasHowarth:
So, I've done some more thinking about why things blow up. Since we have to do some portion of the solve explicitly rather than lagged, I think we revert back to the point I made in the [discussion](https://github.com/AMReX-Combustion/PeleLMeX/issues/389#issuecomment-2176358538rl), and the Y coefficient of the Wbar term becomes a problem. As I say, I can either disable the Wbar corrections in this case and everything works, otherwise this is going to become even more complicated...

ThomasHowarth:
@baperry2 I have resorted to disabling wbar for this particular case - the boundary condition becomes very complicated otherwise. Some of the machinery is still left (e.g. filling boundaries for wbar and computing gradients of wbar using gradients of Y), but it should be bypassed. I'm leaving it there in case someone wants to try and figure this out down the line - I can also remove it. This way the solver is both conservative and consistent, even if the diffusion model isn't perfect. 

Given the number of commits, and switch arounds, it is probably best to squash this merge.

baperry2 (approved):
This plan sounds good. Agree on leaving the wbar code for isothermal Soret in in case we want to revisit. Thanks again for doing this.

Good point on squashing and merging. We actually have GitHub rules set to require squashing and merging for all PRs now!

baperry2:
@jrood-nrel any idea why the `Docs` workflow is not running? It seems to be not running since you updated the workflow ubuntu version.

This doesn't touch Docs, so i'm going to merge, but we should figure that out since it appears to be affecting all PRs

jrood-nrel:
It requires certain files to have changed. I assume that's it. https://github.com/AMReX-Combustion/PeleLMeX/blob/b3d27d4d40943772bdd47a9be8335d91ff9f73a5/.github/workflows/docs.yml#L7-L9 . I would also be fine having it always run.

baperry2:
I think this is fine, I'll just remove it from the list of "required" checks then, otherwise it will show up as "expected" and prevent auto-merging for cases where it does not run.

PR #398, opened 2024-07-16
User function to patch flow variables after restart from plot file

SreejithNREL:
Added a function call so that the user can ignite the flow after reading a plot file.

### DIFF
--- a/Docs/sphinx/manual/Tutorials_BFSFlame.rst
+++ b/Docs/sphinx/manual/Tutorials_BFSFlame.rst
@@ -129,6 +129,9 @@ The user parameters are gathered in the struct defined in  ``pelelmex_prob_parm.
         amrex::Real T_hot  = 1800.0_rt;
         amrex::Real Twall  = 300.0_rt;
         amrex::Real meanFlowMag = 0.0;
+        
+        bool ignite_flow = false;
+        amrex::Real ignition_temperature = 1700.0;
     };
 
 * ``T_mean`` : inlet and initial gas temperature
@@ -145,10 +148,17 @@ The user parameters are gathered in the struct defined in  ``pelelmex_prob_parm.
 
 * ``meanFlowMag`` : inlet :math:`x` velocity
 
+* ``ignite_flow`` : enable ignition after reading a plot solution file
+
+* ``ignition_temperature`` :  The temperature of the ignition kernel. The user is expected to include other ignition parameters. 
+
+
+
 The initial solution consists of a premixed methane/air mixture in the upper part of the domain
 and pure hot air in the wake of the step. The default parameters provided above are overwritten
 using AMReX ParmParse in ``pelelmex_prob.cpp`` and the initial/boundary conditions implemented in
-``pelelmex_prob.H``.
+``pelelmex_prob.H``. Alternatively, the user can write a custom ignition kernel patch function in the ``Patch_Ignition_Source`` function in pelelmex_prob.cpp file. 
+This will patch the ignition kernel conditions after reading a plot solution file.
 
 In addition to these three C++ files, an extra header is needed in the present case compared to
 :doc:`Tutorials_FlameSheet` : ``EBUserDefined.H``. This file is necessary to specify more complex EB
--- a/Docs/sphinx/manual/Tutorials_BFSFlame.rst
+++ b/Docs/sphinx/manual/Tutorials_BFSFlame.rst
@@ -129,9 +129,7 @@ The user parameters are gathered in the struct defined in  ``pelelmex_prob_parm.
         amrex::Real T_hot  = 1800.0_rt;
         amrex::Real Twall  = 300.0_rt;
         amrex::Real meanFlowMag = 0.0;
-        
-        bool ignite_flow = false;
-        amrex::Real ignition_temperature = 1700.0;
+
     };
 
 * ``T_mean`` : inlet and initial gas temperature
@@ -148,17 +146,12 @@ The user parameters are gathered in the struct defined in  ``pelelmex_prob_parm.
 
 * ``meanFlowMag`` : inlet :math:`x` velocity
 
-* ``ignite_flow`` : enable ignition after reading a plot solution file
-
-* ``ignition_temperature`` :  The temperature of the ignition kernel. The user is expected to include other ignition parameters. 
-
-
 
 The initial solution consists of a premixed methane/air mixture in the upper part of the domain
 and pure hot air in the wake of the step. The default parameters provided above are overwritten
 using AMReX ParmParse in ``pelelmex_prob.cpp`` and the initial/boundary conditions implemented in
-``pelelmex_prob.H``. Alternatively, the user can write a custom ignition kernel patch function in the ``Patch_Ignition_Source`` function in pelelmex_prob.cpp file. 
-This will patch the ignition kernel conditions after reading a plot solution file.
+``pelelmex_prob.H``. Alternatively, the user can write a custom function to enforce an ignition kernel through the ``patchFlowVariables`` function in the problem-specific ``PeleLMeX_PatchFlowVariables.cpp`` file. 
+It should be kept in mind that the ``patchFlowVariables`` function can be used if the user wants to patch certain flow variables after reading an existing solution from a plot file ( ``peleLM.do_patch_flow_variables`` should be set to true).
 
 In addition to these three C++ files, an extra header is needed in the present case compared to
 :doc:`Tutorials_FlameSheet` : ``EBUserDefined.H``. This file is necessary to specify more complex EB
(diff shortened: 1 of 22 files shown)


Source/PeleLMeX_Init.cpp
```diff
@@ -38,8 +38,7 @@ PeleLM::MakeNewLevelFromScratch(
       auto const dx = geom[lev].CellSizeArray();
       Real vol = AMREX_D_TERM(dx[0], *dx[1], *dx[2]);
       amrex::Print() << " with " << ba.numPts() << " cells," << ba.size()
-                     << " boxes,"
-                     << " over "
+                     << " boxes," << " over "
```
baperry2: There's probably a slight version mismatch between the clang-format version you are using and that used in the CI that is causing issues. Just don't commit this file.

Docs/sphinx/manual/Tutorials_BFSFlame.rst
```diff
@@ -145,10 +146,12 @@ The user parameters are gathered in the struct defined in  ``pelelmex_prob_parm.
 
 * ``meanFlowMag`` : inlet :math:`x` velocity
 
+
 The initial solution consists of a premixed methane/air mixture in the upper part of the domain
 and pure hot air in the wake of the step. The default parameters provided above are overwritten
 using AMReX ParmParse in ``pelelmex_prob.cpp`` and the initial/boundary conditions implemented in
-``pelelmex_prob.H``.
+``pelelmex_prob.H``. Alternatively, the user can write a custom function to enforce an ignition kernel through the ``patchFlowVariables`` function in the problem-specific ``PeleLMeX_PatchFlowVariables.cpp`` file. 
+It should be kept in mind that the ``patchFlowVariables`` function can be used if the user wants to patch certain flow variables after reading an existing solution from a plot file ( ``peleLM.do_patch_flow_variables`` should be set to true).
```
baperry2, line 154: Can you also add `peleLM.do_patch_flow_variables` to the controls section of the documentation?

Exec/RegTests/EB_BackwardStepFlame/input.2d
```diff
@@ -69,7 +69,7 @@ peleLM.progressVariable.coldState = CO:0.0 CO2:0.0
 peleLM.progressVariable.hotState = CO:0.000002 CO2:0.0666
 
 #---------------------- Reactor CONTROL --------------------------
-peleLM.chem_integrator = "ReactorCvode"
+peleLM.chem_integrator = "ReactorNull"
```
baperry2: Don't change this so that the tutorial functions as expected.

Exec/RegTests/EB_BackwardStepFlame/input.2d
```diff
@@ -29,14 +29,13 @@ prob.T_wall = 300
...
 prob.Y_oxid = 0.2226345
-
 #---------------------- PeleLM CONTROL ---------------------------
 peleLM.v = 1                                # PeleLMeX verbose
 peleLM.use_wbar = 1                         # Include Wbar term in species diffusion fluxes
 peleLM.sdc_iterMax = 2                      # Number of SDC iterations
 peleLM.num_init_iter = 3                    # Number of initial iterations
 peleLM.floor_species = 0                    # Enforce species positivity (non-conservative !)
-
+peleLM.do_patch_flow_variables = true
```
baperry2: set to false because this isn't actually used in the tutorial.

Source/PeleLMeX_Plot.cpp
```diff
@@ -951,50 +952,58 @@ PeleLM::initLevelDataFromPlt(int a_lev, const std::string& a_dataPltFile)
...
   ProbParm const* lprobparm = prob_parm_d;
 
-  // Enforce rho and rhoH consistent with temperature and mixture
-  // The above handles species mapping (to some extent), but nothing enforce
-  // sum of Ys = 1 -> use N2 in the following if N2 is present
+  // If m_do_patch_flow_variables is set as true, call user-defined function to
+  // patch flow variables
+  if (m_do_patch_flow_variables) {
+    patchFlowVariables(geom[a_lev], *lprobparm, ldata_p->state);
+  } else {
```
baperry2: Should this be en `else`? I feel like the correction below should happen even when patchFlowVariables is being used. That would allow the user to be "lazy" and for exmaple just redefine temperature and count on the logic below to appropriately reset enthalpy.

Exec/RegTests/EB_BackwardStepFlame/GNUmakefile
```diff
@@ -27,6 +27,7 @@ THREAD_SANITIZER = FALSE
 
 # PeleLMeX
 CEXE_headers += EBUserDefined.H
+#CEXE_sources += PeleLMeX_PatchFlowVariables.cpp
```
baperry2: Delete if not needed

Source/PeleLMeX_Timestep.cpp
```diff
@@ -23,10 +23,12 @@ PeleLM::computeDt(int is_init, const TimeStamp& a_time)
       estdt = m_init_dt;
     } else {
       Real dtconv = estConvectiveDt(a_time);
+      amrex::Print() << "\nConvective DT = " << dtconv;
```
baperry2: Do we need to add these? The convective and divergence dt already get printed.

Source/PeleLMeX_PatchFlowVariables.H
```diff
@@ -0,0 +1,10 @@
+#ifndef PELELM_PATCHFLOWVARIABLES_H
+#define PELELM_PATCHFLOWVARIABLES_H
+
+class PeleLM;
+
+void patchFlowVariables(
```
baperry2, line 6: How about making this a member of the PeleLM class (move the declaration to PeleLMeX.H)?

Source/Make.package
```diff
@@ -43,6 +44,7 @@ CEXE_sources += PeleLMeX_Diagnostics.cpp
 CEXE_sources += PeleLMeX_FlowController.cpp
 CEXE_sources += PeleLMeX_DeriveUserDefined.cpp
 CEXE_sources += PeleLMeX_BPatch.cpp
+CEXE_sources += PeleLMeX_PatchFlowVariables.cpp
```
baperry2, line 47: You'll also have to modify `CMake/BuildPeleExe.cmake` fof the CMake build system to work.

baperry2 (changes requested):

Source/PeleLMeX_Init.cpp
SreejithNREL: change incorporated

Exec/RegTests/EB_BackwardStepFlame/input.2d
SreejithNREL: change incorporated

Source/PeleLMeX_Plot.cpp
SreejithNREL: yes, you are right. it is better to keep it that way.
change incorporated

Exec/RegTests/EB_BackwardStepFlame/GNUmakefile
SreejithNREL: Done

Source/Make.package
SreejithNREL: Done

Source/PeleLMeX_PatchFlowVariables.H
SreejithNREL: Let's consider this for a future PR if it's not mandatory now?

baperry2: Sure, sounds good.

baperry2 (approved):

Exec/RegTests/EB_BackwardStepFlame/PeleLMeX_PatchFlowVariables.cpp
```diff
@@ -0,0 +1,32 @@
...
+    auto const& rhoY_arr = a_mf.array(mfi, FIRSTSPEC);
+    auto const& rhoH_arr = a_mf.array(mfi, RHOH);
+    auto const& temp_arr = a_mf.array(mfi, TEMP);
+    amrex::ParallelFor(bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+      amrex::Real x[3] = {
+        prob_lo[0] + static_cast<amrex::Real>(i + 0.5) * dx[0],
+        prob_lo[1] + static_cast<amrex::Real>(j + 0.5) * dx[1],
+        prob_lo[2] + static_cast<amrex::Real>(k + 0.5) * dx[2]};
+
+      /*User can define how to patch flow variables here.*/
```
baperry2, line 25: You'll have to comment out or delete unused variables for this to pass tests. But actually the best way would be to populate this with a simple sample function, e.g. setting the temperature to a specific value over some region, so that these aren't unused variables.

PR #401, opened 2024-07-23
New Options for Chi Correction in MAC Projections

baperry2:
This was needed for the manifold stuff, so moving it in now to capture some progress toward fully merging that capability. Probably won't merge the rest for a while though.

This PR does two things:
- Optionally print the zero norm of the chi correction on each SDC iteration, which indicates the mismatch between the actual pressure and the prescribed pressure based on the Low Mach assumption. This can be used to assess how the SDC iterations are converging. Note that the values printed are the mismatch before the indicated iteration, so the value for the 0th iteration on one time step corresponds to the value at the end of the previous timestep. That output looks something like this:
```
   SDC iter [1] 
      Before SDC 1: max relative P mismatch is 0.0002924937001
   - oneSDC()::MACProjection()   --> Time: 0.003672
   - oneSDC()::ScalarAdvection() --> Time: 0.009108
   - oneSDC()::ScalarDiffusion() --> Time: 0.059437
   - oneSDC()::ScalarReaction()  --> Time: 0.76213
   SDC iter [2] 
   - oneSDC()::Update t^{n+1,k}  --> Time: 0.029643
      Before SDC 2: max relative P mismatch is 0.01296565691
   - oneSDC()::MACProjection()   --> Time: 0.003565
   - oneSDC()::ScalarAdvection() --> Time: 0.009067
   - oneSDC()::ScalarDiffusion() --> Time: 0.057344
   - oneSDC()::ScalarReaction()  --> Time: 0.757911
```
- Offer new strategies for how the divergence constraint on each SDC iteration is computed: 
   - `DivuEveryIter` (current, remains default), 
   - `DivuFirstIter` (the estimate of Divu computed on the 0th SDC iteration is kept constant, and only the chi correction is used to drive the solution toward convergence)
   - `NoDivu` (the Divu estimates are ignored in the MAC projection and only the chi correction is used to iteratively ensure the actual pressure (RhoRT) matches the desired value)

For the PMF case, it can be observed that the second option provides better convergence of the pressure than the first. My working hypothesis is that updating the estimated divergence constraint essentially adds noise to the iterative update of the chi correction. The converged solution with infinite SDC iterations would be the same between all methods, but the rate of convergence is not. The existing approach is left as the default because the alternative `DivuFirstIter` approach hasn't been tested for a wide range of cases yet. The `NoDivu` option should never be used and is only included for comparison. In the latter two options, some compute could be saved by not recomputing the divergence estimate at each n+1,k, but for now it is computed but then ignored.

SDC-convergence

### DIFF
--- a/Docs/sphinx/manual/LMeXControls.rst
+++ b/Docs/sphinx/manual/LMeXControls.rst
@@ -123,7 +123,7 @@ IO parameters
     amr.initDataPlt      = plt01000        # [OPT, DEF=""] Provide a plotfile from which to extract initial data
     amr.regrid_on_restart = 1              # [OPT, DEF="0"] Trigger a regrid after the data from checkpoint are loaded
     amr.n_files          = 64              # [OPT, DEF="min(256,NProcs)"] Number of files to write per level
-    
+
 Refinement controls
 -------------------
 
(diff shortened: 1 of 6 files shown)


drummerdoc (approved):
Not surprising that divu can be a bit messy.  Does this get better with resolution?

baperry2:
The convergence gets much better with smaller timesteps. I'm not sure about grid resolution, but I imagine that also helps.

PR #402, opened 2024-07-26
Reset time and nsteps

SreejithNREL:
Capability to reset (or not) time and nsteps after restarting from a plot file

### DIFF
--- a/Docs/sphinx/manual/LMeXControls.rst
+++ b/Docs/sphinx/manual/LMeXControls.rst
@@ -121,6 +121,7 @@ IO parameters
 
     amr.restart          = chk00100        # [OPT, DEF=""] Checkpoint from which to restart the simulation
     amr.initDataPlt      = plt01000        # [OPT, DEF=""] Provide a plotfile from which to extract initial data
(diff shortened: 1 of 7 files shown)


Docs/sphinx/manual/LMeXControls.rst
```diff
@@ -256,6 +256,7 @@ PeleLMeX algorithm
     peleLM.deltaT_tol = 1e-10              # [OPT, DEF=1.e-10] Tolerance of the deltaT solve
     peleLM.evaluate_vars =...              # [OPT, DEF=""] In evaluate mode, list unitTest: diffTerm, divU, instRR, transportCC
     peleLM.do_patch_flow_variables = false # [OPT, DEF=false] Enable user-defined flow variable patching after reading a plot solution file
+    peleLM.do_reset_time = 1               # [OPT, DEF=1] Resets time and nsteps to 0 after restarting from a plot file. (Warning: plot file will be rewritten if not renamed and argument value = 0)
```
baperry2: How about calling this `peleLM.initDataPlt_reset_time` so that it's obvious this setting only applies when initializing from a plt file? And a similar change for `peleLM.do_path_flow_variables`. Also, let's move these control entries to immediately below `amr.initDataPlt` in the documentation to make the association even more clear.

baperry2 (approved):
Looks good other than the minor change requested, so just go ahead and merge after making that change.

PR #404, opened 2024-08-12
Spark ignition utility

ThomasHowarth:
This PR adds the capability to add a source term to the energy equation, termed "spark".  The energy is deposited in a discrete spherical region at a particular time, for a specified duration - some discussion could be had about making these continuous functions, rather than being clipped in this way. Multiple sparks can be defined, and inputs have been added to documentation. Addresses some discussions had [here](https://github.com/AMReX-Combustion/PeleLMeX/discussions/384).

### DIFF
--- a/Source/PeleLMeX_Advance.cpp
+++ b/Source/PeleLMeX_Advance.cpp
@@ -74,6 +74,9 @@ PeleLM::Advance(int is_initIter)
 
   for (int lev = 0; lev <= finest_level; lev++) {
     m_extSource[lev]->setVal(0.);
+    if (m_n_sparks > 0) {
+      addSpark(lev, AmrNewTime);
+    }
   }
   //----------------------------------------------------------------
 
@@ -109,6 +112,12 @@ PeleLM::Advance(int is_initIter)
   //----------------------------------------------------------------
   BL_PROFILE_VAR_STOP(PLM_SETUP);
   //----------------------------------------------------------------
+
+  if (m_n_sparks > 0) {
+    for (int lev = 0; lev <= finest_level; lev++) {
+      addSpark(lev, AmrOldTime);
+    }
+  }
 
 #ifdef PELE_USE_SPRAY
   if (is_initIter == 0) {
--- a/Source/PeleLMeX_Advance.cpp
+++ b/Source/PeleLMeX_Advance.cpp
@@ -74,9 +74,6 @@ PeleLM::Advance(int is_initIter)
 
   for (int lev = 0; lev <= finest_level; lev++) {
     m_extSource[lev]->setVal(0.);
-    if (m_n_sparks > 0) {
-      addSpark(lev, AmrNewTime);
-    }
   }
   //----------------------------------------------------------------
 
@@ -114,9 +111,7 @@ PeleLM::Advance(int is_initIter)
   //----------------------------------------------------------------
 
   if (m_n_sparks > 0) {
-    for (int lev = 0; lev <= finest_level; lev++) {
-      addSpark(lev, AmrOldTime);
-    }
+    addSpark(AmrOldTime);
   }
 
 #ifdef PELE_USE_SPRAY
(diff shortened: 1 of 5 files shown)


Source/PeleLMeX_Forces.cpp
```diff
@@ -171,3 +171,70 @@ PeleLM::getVelForces(
...
+                 rho_a, rhoY_a, temp_src_a, rhoh_src_a, eos,
+                 dx] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+          Real dist_to_center = std::sqrt(AMREX_D_TERM(
+            (i - spark_idx[0]) * (i - spark_idx[0]) * dx[0] * dx[0],
+            +(j - spark_idx[1]) * (j - spark_idx[1]) * dx[1] * dx[1],
+            +(k - spark_idx[2]) * (k - spark_idx[2]) * dx[2] * dx[2]));
+          if (dist_to_center < spark_radius) {
+            // probably doesn't actually
+            // contribute to anything
+            temp_src_a(i, j, k) = spark_temp / spark_duration;
```
baperry2: Yes, this is not needed so you can remove

Source/PeleLMeX_Forces.cpp
```diff
@@ -171,3 +171,70 @@ PeleLM::getVelForces(
...
+    }
+    Box domainBox = geom[lev].Domain();
+    // just a check
+    if (!domainBox.contains(spark_idx)) {
+      Warning(m_spark[n] + " not in domain!");
+      continue;
+    }
+    auto* ldata_p = getLevelDataPtr(lev, a_timestamp);
+    auto eos = pele::physics::PhysicsType::eos();
+    for (MFIter mfi(*(m_extSource[lev]), TilingIfNotGPU()); mfi.isValid();
```
baperry2: Add OMP pragma similar to other ParallelFors in this file:
```
#ifdef AMREX_USE_OMP
#pragma omp parallel if (Gpu::notInLaunchRegion())
#endif
```

Or, the better option would be to convert this into the newer version of ParallelFor that operates directly on a MultiFab without MFIter (see, e.g, https://github.com/AMReX-Combustion/PeleLMeX/blob/8dccce90dfba3a1000a1c1602e701a06a0d09b7b/Source/PeleLMeX_Advection.cpp#L749, and note the `Gpu::StreamSynchronize` at the end of the loop)

Source/PeleLMeX_Advance.cpp
```diff
@@ -74,6 +74,9 @@ PeleLM::Advance(int is_initIter)
 
   for (int lev = 0; lev <= finest_level; lev++) {
     m_extSource[lev]->setVal(0.);
+    if (m_n_sparks > 0) {
+      addSpark(lev, AmrNewTime);
```
baperry2: Is this needed? We don't have an equivalent source term call for soot/spray/rad.

baperry2 (changes requested):
Mostly looks good other than the few comments I added.

Source/PeleLMeX_Forces.cpp
ThomasHowarth: I've made this change - I assume there's a lot of functions lying around that need this change...

Source/PeleLMeX_Advance.cpp
ThomasHowarth: Doesn't look like it - removed.

Source/PeleLMeX_Forces.cpp
baperry2: Yeah there probably are. The multifab level ParFor can be a bit more efficient on GPU and simplifies the code (not needing MFiter and OMP pragmas), but it isn't a big enough deal to be a priority to go back and switch the old ones. Lucas switched a bunch of them over quite a while ago in potentially higher impact places.

baperry2 (approved):

PR #410, opened 2024-09-04
Print temporals in CSV format

baperry2:
Add headers with variable names to temporals and convert to comma separated rather than space separated format for easier processing.

Add a function to easily get the state variable name from its index.


Source/PeleLMeX_Utils.cpp
SreejithNREL, line 1: Line No: 1213, in the print statement, it should be "invalid" and not "invalod".

SreejithNREL (changes requested):
Please correct the typo and otherwise the changes look good.

SreejithNREL (approved):

PR #414, opened 2024-09-13
Eosparm support in PeleLMeX

baperry2:
Progress toward merging the manifold modeling stuff into PeleLMeX

Pass eosparm as argument into all EOS functions. Right now this is not necessary because it is only used by Manifold EOS, but this paves the way for that and should not negatively impact anything. 
- Does not yet support all Efield functions (but because eosparm is not needed for non-manofold EOS, this does not impact functionality)
- FlameSheet case shows example of how eosparm can be used in problem files for bcnormal/initdata. This will have to be added to any case to be run with manifold EOS, but is done in a way that does not break existing cases.

### DIFF
--- a/Source/PeleLMeX_DeriveFunc.cpp
+++ b/Source/PeleLMeX_DeriveFunc.cpp
@@ -75,8 +75,9 @@ pelelmex_derheatrelease(
   auto const react = reactfab.const_array(0);
   auto const& Hi = EnthFab.array();
   auto HRR = derfab.array(dcomp);
+  auto const* leosparm = a_pelelm->eos_parms.device_parm();
   amrex::ParallelFor(bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
-    getHGivenT(i, j, k, temp, Hi);
+    getHGivenT(i, j, k, temp, Hi, leosparm);
     HRR(i, j, k) = 0.0;
     for (int n = 0; n < NUM_SPECIES; n++) {
       HRR(i, j, k) -= Hi(i, j, k, n) * react(i, j, k, n);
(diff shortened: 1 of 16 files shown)


Source/PeleLMeX_DeriveFunc.cpp
```diff
@@ -1418,15 +1421,16 @@ pelelmex_derlambda(
...
   auto rhotheta = dummies.array(NUM_SPECIES + 1);
   auto const* ltransparm = a_pelelm->trans_parms.device_parm();
+  auto const* leosparm = a_pelelm->eos_parms.device_parm();
   amrex::Real LeInv = a_pelelm->m_Lewis_inv;
   amrex::Real PrInv = a_pelelm->m_Prandtl_inv;
   amrex::ParallelFor(
-    bx,
-    [do_fixed_Le, do_fixed_Pr, do_soret, LeInv, PrInv, rhoY, T, rhoD, rhotheta,
-     lambda, mu, ltransparm] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+    bx, [do_fixed_Le, do_fixed_Pr, do_soret, LeInv, PrInv, rhoY, T, rhoD,
```
marchdf, line 1428: In general, it is interesting that lmex does the capture explicitly. In C we let the compiler decide with `=`.

marchdf (approved):

Source/PeleLMeX_DeriveFunc.cpp
baperry2: I think PeleLMeX is inconsistent about which way it does it. Being explicit can be more clear/helpful, but it does add a bit of annoyance in the code. So I just followed what was there.

PR #415, opened 2024-09-18
Add full manifold model capability

baperry2:
Depends on #414


baperry2:
This is now mostly ready, except for documentation and adding to the CI. 

There is a FlameSheet example case (`input.manifold`) that can be run by compiling with `USE_MANIFOLD = TRUE`. That case requires a pmf data file and a tabulated chemistry file that can be run using the script `get_pmfdat_and_chemtable.py` in the CMLM repo: https://github.com/NREL/cmlm/blob/development/get_pmfdat_and_ctable.py

The test case uses a 1D table based on cantera premixed flames at the same conditions as the detailed chemistry FlameSheet test.

Source/PeleLMeX_Setup.cpp
```diff
@@ -459,6 +467,11 @@ PeleLM::readParameters()
                       "fixed_Pr or fixed_Le is true"
                    << std::endl;
   }
+  if (
+    m_use_wbar != 0 &&
```
marchdf: add extra paren?

Source/PeleLMeX_Setup.cpp
```diff
@@ -1118,6 +1131,20 @@ PeleLM::derivedSetup()
     "enstrophy", IndexType::TheCellType(), 1, pelelmex_derenstrophy,
     grow_box_by_two);
 
+#ifdef USE_MANIFOLD_EOS
+  auto& mani_data = eos_parms.host_only_parm().manfunc_par->host_parm();
+  int nmanivar = mani_data.Nvar;
```
marchdf:
```suggestion
  const int nmanivar = mani_data.Nvar;
```

Source/PeleLMeX_K.H
```diff
@@ -625,7 +735,8 @@ makeVelForce(
   const amrex::Real& rho_incomp,
   int pseudo_gravity,
   int pseudo_gravity_dir,
-  const amrex::Real& /*time*/,
+  const amrex::Real& /*time*/
+  ,
```
marchdf: this is a strange diff?

marchdf (approved):
Left some minor comments

Exec/Make.PeleLMeX
```diff
@@ -73,6 +73,20 @@ endif
 ifeq ($(Eos_Model),$(filter $(Eos_Model),Soave-Redlich-Kwong))
    DEFINES += -DUSE_SRK_EOS
 endif
+ifeq ($(Eos_Model),$(filter $(Eos_Model),Manifold))
+   DEFINES += -DUSE_MANIFOLD_EOS
+   DEFINES += -DUSE_MANIFOLD_EOS
```
drummerdoc: why twice?

Exec/Make.PeleLMeX
```diff
@@ -73,6 +73,20 @@ endif
...
+   DEFINES += -DMANIFOLD_DIM=$(Manifold_Dim)
+   ifeq ($(Manifold_Type),Table)
+      DEFINES += -DMANIFOLD_EOS_TYPE=1
+   else
+      ifeq ($(Manifold_Type),Network)
+         DEFINES += -DMANIFOLD_EOS_TYPE=2
+      else
+         DEFINES += -DMANIFOLD_EOS_TYPE=0
```
drummerdoc, line 89: Wonder if, later, these can be words rather than integers.

drummerdoc (approved):

Exec/Make.PeleLMeX
baperry2: fixed.

Exec/Make.PeleLMeX
baperry2: At least the user-facing part is words for now.

baperry2:
Going to merge this now. Will add docs and more testing in future PRs.

PR #417, opened 2024-09-23
Parse for adv_redist_type and diff_redist_type

d-montgomery:
Added the ability to change the redistribution type for advection and diffusion by parsing input files.

### DIFF
--- a/Docs/sphinx/manual/LMeXControls.rst
+++ b/Docs/sphinx/manual/LMeXControls.rst
@@ -393,6 +393,12 @@ in ``Exec/RegTest/EB_BackwardStepFlame`` and ``Exec/RegTest/EB_FlowPastCylinder`
 .. note::
    Note that when using isothermal EB in combination with LES, the thermal diffusion coefficient employed to compute the EB boundary thermal flux only uses the molecular contribution.
 
+Lastly, it is possible to change the default redistribution scheme described in the :ref:`geometry with embedded boundaries section: <ssec:geoEB>`
+::
+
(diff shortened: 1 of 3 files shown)


baperry2 (approved):
Thanks for adding this!

baperry2:
For posterity sake to remember the purpose of this: with isothermal EB, or potentially other EB cases, the chi_correction in the SDC iterations was not converging. This was because the redistribution process is arbitrary and does not enforce that RhoRT is maintained in the redistribution. We do a lot of work to ensure that the MAC velocity field divergence will result in constant RhoRT, then throw that away by applying redistribution to scalar advection. @dmontgomeryNREL found that turning off advection redistribution resolved our problems and did not lead to instability.

d-montgomery:
Case setup:
Rectangular domain with isothermal cylinder EB initialized with pure N2 at 143 K.  The inlet CO2 mass fraction and temperature slowly increase from $Y_{CO2} = 0 \rightarrow 1$ and $T = 143 \rightarrow 293$ K, while the inlet N2 mass fraction decreases from $Y_{N2} = 1 \rightarrow 0$.  

test_2Spec_EB_3SDC_DivuFirstIter_StateRedist_Surf

These plots demonstrate the effect of turning off flux redistribution for advection.  The first plot is from a divergent simulation that resulted in NaN's with advective fluxes treated with state redistribution.  The second is from a simulation where the redistribution is turned off for the advective fluxes. 

test_2Spec_EB_3SDC_DivuFirstIter_StateRedist

test_2Spec_EB_3SDC_DivuFirstIter_NoRedist

PR #427, opened 2024-10-15
"ODE" quantities and user defined external sources

d-montgomery:
This PR introduces a quantity that varies in time, but is neither advected or diffused.  I'll refer to it as an `ode_qty` for now.  In PeleC these would be `AUX` quantities, but there already exists a placeholder for `AUX` in PeleLMeX that is not a state variable.  I intend to rename the existing aux quantities to `ADV` and these ode quantities to `AUX` for consistency with PeleC, but that will be a large PR in itself.

The ode quantities are included in the state variable and satisfy $\frac{\partial B_k}{\partial t} = S_{ext,B_k}$, for $k = 0, 1,...,$NUM_ODE.  Currently this is only first order accurate in time, but is consistent with the treatment of external sources from soot, spray and radiation, which are calculated prior to the SDC iterations.  In the future I would like to add the ability to update the source terms within the SDC iterations, but this may require solving additional nonlinear systems depending on the source terms. For now we assume that the source terms are constant from $t^{n} \rightarrow t^{n+1}$.

In addition to the ode quantities, this PR includes `PeleLMeX_ProblemSpecificFunctions.cpp` which allows users to define the name of the ode quantities as well as external source terms for any state variable.  Note that external source terms for density, $S_{ext,\rho}$, affect the velocity constraint but this has not been addressed in this PR.  For reference, the velocity constraint should be: 
$\nabla \cdot \vec{u} = \frac{1}{T}\frac{DT}{Dt}  + W \displaystyle\sum_m \frac{1}{W_m} \frac{DY_m}{Dt} + \frac{1}{\rho} S_{\text{ext},\rho} = S.$
Note that the existing  external force terms for soot, spray and radiation are now  included in the function called `getExternalSources` in `PeleLMeX_Forces.cpp`.

I tested 3 ode quantities of increasing stiffness with $S_{ext} = -10^{k+2} \cdot B_k$, for $k = 0,1,2$.  The results below demonstrate first order accuracy and the need for SDC iterations in the future.
 
ODEQty

solutions

convergence_SDC-0_stiff


Docs/sphinx/manual/Model.rst
```diff
@@ -118,6 +118,12 @@ For the standard ideal gas EOS, the divergence constraint on velocity becomes:
     \nabla \cdot \boldsymbol{u} &= \frac{1}{\rho c_p T} \left(\nabla \cdot \lambda\nabla T - \sum_m \boldsymbol{\Gamma_m} \cdot \nabla h_m \right) \\
     &- \frac{1}{\rho} \sum_m \frac{W}{W_m} \nabla \cdot \boldsymbol{\Gamma_m} + \frac{1}{\rho}\sum_m \left(\frac{W}{W_m} - \frac{h_m}{c_p T} \right) \dot \omega \equiv S .
 
+In addition to the flow equations, `PeleLMeX` can also solve for a set of quantities that are neither advected nor diffused, satisfying:
+
+.. math::
+
+    \frac{\partial B_k}{\partial t} = S_{\text{ext},B_k}.
```
baperry2, line 150: Should be `d` rather than `\partial` because these only change with time?

Source/PeleLMeX_Plot.cpp
```diff
@@ -281,6 +281,14 @@ PeleLM::WritePlotFile()
     plt_VarsName.push_back("viscturb");
   }
 
+#if NUM_ODE > 0
+  for (int n = 0; n < NUM_ODE; n++) {
+    plt_VarsName.push_back(m_ode_names[n]);
+  }
+#endif
+
+  Print() << PrettyLine;
```
baperry2: Remove print statement

Source/PeleLMeX_Plot.cpp
```diff
@@ -401,6 +409,12 @@ PeleLM::WritePlotFile()
       cnt += m_ionsFluxes[lev]->nComp();
     }
 #endif
+#if NUM_ODE > 0
+    MultiFab::Copy(
+      mf_plt[lev], m_leveldata_new[lev]->state, FIRSTODE, cnt, NUM_ODE, 0);
+    cnt += NUM_ODE;
+    Print() << "cnt = " << cnt << std::endl;
```
baperry2: remove print statement

baperry2 (changes requested):
Looks good. I'm able to run the test case and get the expected results. Just a couple minor changes requested. 

One slightly larger thing is to have this run in the CI. You'll need to add it in `Tests/CMakeLists.txt` and you'll probably also need to add something to define NUM_ODE with the CMake build system similar to NUM_AUX in PeleC: https://github.com/AMReX-Combustion/PeleC/blob/b9ebe80bd3b00db0f50710048e50bac01212ef2b/CMake/BuildPeleExe.cmake#L25

And a clarification is that these quantities aren't spatially invariant, but don't advect or diffuse, so spatial terms don't come up in their governing equations.

Docs/sphinx/manual/Model.rst
d-montgomery: $B_k = B_k(x_j,t)$ so the \partial is needed.

Tests/CMakeLists.txt
```diff
@@ -225,3 +225,11 @@ if(NOT PELE_ENABLE_EB)
     add_test_r(hit-les-${PELE_DIM}d HITDecay)
   endif()
 endif()
+
+if (PELE_ENABLE_EB)
+  if(PELE_DIM EQUAL 2)
+    if(PELELM_NUM_ODE GREATER 0)
```
baperry2: It looks like this test isn't actually running. I don't think we need the `if(PELELM_NUM_ODE GREATER 0)` - this gets accounted for in the CMakeLIsts.txt for the EB_ODEQty case

d-montgomery: You're absolutely right.  I've addressed this locally and will push a commit soon.  There were some other issues related to the `ProblemSpecifcFunctions` too.

baperry2 (approved):

PR #435, opened 2024-11-06
Update PelePhysics and add path to new utility 

d-montgomery:
Pulling in the most recent PelePhysics commit and updating the paths in Make.PeleLMeX.


baperry2:
Can you replace some of the unit conversions (e.g., [here](https://github.com/AMReX-Combustion/PeleLMeX/blob/86c4721801ad8b784d83e0fa17e1e6d55abfbb54/Source/PeleLMeX_K.H#L50)) with this capability as demonstration and to ensure it gets caught in some of our tests?

baperry2 (approved):

PR #437, opened 2024-11-10
Use ubuntu-24.04 in CI

jrood-nrel:


Tests/CMakeLists.txt
```diff
@@ -210,10 +219,10 @@ endfunction(add_test_spray)
 #=============================================================================
 
 if(NOT PELE_ENABLE_EB)
-  add_test_r(flamesheet-drm19-${PELE_DIM}d FlameSheet)
+  add_test_re(flamesheet-drm19-${PELE_DIM}d FlameSheet)
```
baperry2: These shouldn't be excluded from the CI. If they're taking too long we should reduce the size instead.

baperry2:
Just a note that after merging we'll need to update the names of the required tests in the branch rules to get rid of the "expected" checks that aren't running, because GitHub isn't smart enough to do that on it's own.

Tests/CMakeLists.txt
jrood-nrel: Ok, maybe I will do what we in PeleC and only run like 1 test under `Debug`.

baperry2 (approved):

PR #446, opened 2024-12-09
Try moving ubuntu CI image to older version

jrood-nrel:
Let's see if this fixes the MPI issues.


jrood-nrel:
Looks like it's working in https://github.com/AMReX-Combustion/PeleLMeX/pull/443

PR #448, opened 2024-12-12
add error check on pressure matching for Manifold EOS

baperry2:
In the Manifold EOS, the nominal pressure for the manifold model must be specified. This pressure should match the Pmean specified for PeleLMeX in the prob structure, although one is in cgs units and the other is in mks units.

### DIFF
--- a/Source/PeleLMeX.H
+++ b/Source/PeleLMeX.H
@@ -316,6 +316,7 @@ public:
    * parameters.
    */
   void checkRunParams();
+  void checkSetupParams();
 
   //-----------------------------------------------------------------------------
   // REGRID / LOAD BALANCE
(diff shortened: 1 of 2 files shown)


baperry2:
@justint1997

SreejithNREL (commented):
Should we let the error tolerance be hard-coded to 1e-6 or let the user choose it?

baperry2:
No need to make this user specifiable as it doesn't affect the solution. It's just there as a sanity check that the user is specifying valid inputs. 1e-6 is plenty to ensure that users don't forget to keep these values the same.

SreejithNREL (approved):
Looks good to me.

PR #451, opened 2024-12-18
Dissipation rate source terms for manifold variable variances

baperry2:
- Modify where turbulent viscosity gets calculated so it is available for dissipation rate modeling
- Add external forcing term with dissipation rate source for variances (as indicated through updates to manifold EOS in PelePhysics)
- Add option to plot external forcings 

Still WIP, needs some debugging


baperry2:
The test case, `input.cvar` in FlameSheet, can be run after using the script and instructions in https://github.com/NREL/cmlm/pull/5 to create the 2D chemtable and then copying the chemtable and PMF data files over the FlameSheet directory.

baperry2:
Test case (`Exec/RegTests/Flamesheet/input.manifold.cvar`) can be run by first executing the `python get_pmfdat_and_ctable_withvar.py` in the [cmlm repo](https://github.com/NREL/cmlm/blob/development/get_pmfdat_and_ctable_withvar.py), then copying the files `data/prem_drm19_phi1_p1_t298.ctb` and `data/prem_drm19_phi1_p1_t298_mani_filter2.344e-03.dat` to the FlameSheet directory. The FlameSheet case needs to be compiled with `USE_MANIFOLD=TRUE` and `Manifold_Dim = 2`.

baperry2:
I've realized that I did not include the production rate in the variance equation (see below). To include this we will need to evaluate scalar gradients, but once we have those it should be very straightforward to add.

Screenshot 2025-03-12 at 10 51 16 AM

baperry2:
Merging this for some work that depends on it, but it should be treated as an "under active development" feature.

PR #453, opened 2024-12-19
fix error for divu calc for manifold when its not being used with Null mech

### LINKED ISSUE #452: manifold compute_divu function incompatible with Null mechanism
With the addition of the manifold capabilities, the compute_divu (when the Manifold EOS) function uses NUM_SPECIES-1 for setting up the [derivs array](https://github.com/AMReX-Combustion/PeleLMeX/blob/652f0eaeadef711653757f66de83ce4a124fd4c9/Source/PeleLMeX_K.H#L348), but this is incompatible with the Null mechanism when compiling without USE_MANIFOLD and MANIFOLD_DIM flag since it only has 1 species. I was just compiling something for an incompressible run  - should this function only be compiled when USE_MANIFOLD is active and use MANIFOLD_DIM instead of NUM_SPECIES - 1? Perhaps @baperry2 is the best person to answer this question.

baperry2:
Closes #452 

@ThomasHowarth: can you verify this works for you?

### DIFF
--- a/Source/PeleLMeX_K.H
+++ b/Source/PeleLMeX_K.H
@@ -345,13 +345,13 @@ compute_divu<pele::physics::eos::Manifold>(
 
   auto eos = pele::physics::PhysicsType::eos(eosparm);
   eos.RY2RRinvY(massdens, rho, rhoinv, y);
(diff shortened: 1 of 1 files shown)


ThomasHowarth:
I can confirm that this fixes the compile problem that I had.

PR #461, opened 2025-01-06
Counterflow Diffusion Flame Validation

justint1997:
Added section to Validation.rst including validation results comparing PeleLMeX counterflow flame centerline profiles to that of Cantera.


baperry2:
Thanks for adding this @justint1997! The documentation additions look great.

Could you also add the input file (and any modifications the the `pelelmex_prob*` files if applicable) that you used to run this case to the `PeleLMeX/Exec/Production/CounterFlow` directory? And if the python script to run the Cantera flame and generate the comparison plots is simple enough, it would be great if you could add that too.

Exec/Production/CounterFlow/plt02656_coolflow/Header
baperry2, line 1: Don't include the plt data as this will bloat the git repository.

justint1997: Okay, I deleted the plt data.

baperry2 (changes requested):
Can you add a README with a short description of the case and how to run it with the cold flow first and then ignition? See an example here: https://github.com/AMReX-Combustion/PeleLMeX/blob/development/Exec/RegTests/FlameSheet/README.md

Then I think this is good.

baperry2 (approved):

SreejithNREL (approved):

Issue #156, opened 2022-12-13
Developments shopping list


esclapez:
- [x] Refactor problem setup #477 
- [x] #473
- [ ] Add on-the-fly diagnostics: PDF, conditional mean, iso-surface, ...
- [ ] Extend CI
- [ ] Extend V&V
- [x] Add more tutorials
- [ ] Machinery to control dt when adding levels
- [ ] Add option to prune covered boxes when EB
- [ ] Time averaged data

baperry2:
@esclapez What's the status of EB-inflow capability? I know you had a branch with this at least partially set up (https://github.com/AMReX-Combustion/PeleLMeX-archive/tree/EBInflow_Updated). Was that working or is more development needed? We have a case we'd like to run that would need EB-inflow capability.

esclapez:
@baperry2 It's been on the back burner mostly because it is pretty invasive, but it is working. @EnnaDelfen is using it on one of her case (brought up to date with LMeX from a few months ago in her fork), even though she would like turbulence on EB-inflow ... always more !
The way it is implemented uses a new `struct` that handles all the EB-related needs (custom geom, EB state and so on). To avoid having to define everything when you have plain EB it uses SFINAE which I'm still on the fence about whether it is a good solution or not. I can try to refresh this, see if I can minimize the APIs breaks and open a draft PR so that we can start a discussion.
On the long run, I think it would be nice gather all the initial/boundary functions and data in a `struct` (maybe common to both C and LMeX ?)

baperry2:
Thanks Lucas! Good to know this is working. I created a branch that rebases the EB-Inflow work to current development over here and verified that the test case still works as expected: https://github.com/baperry2/PeleLMeX/tree/ebinflow-rebase2

I think it would be good to get this merged in but having the up to date branch is enough for us to get started on the project that will use this capability. If you have a bit of time to make any modifications you'd like and create a draft PR, that would be awesome. Otherwise, I'll try to work toward getting this merged at some point.

I agree with the goal of eventually pushing all the problem-dependent BC/IC stuff into a single `struct`. We moved in that direction slightly in that direction for PeleC in https://github.com/AMReX-Combustion/PeleC/pull/743, but there's still a lot to do there. I think the codes are different enough that they couldn't share this struct, but each code have their own.

baperry2:
Just messing around a bit here with github's new sub-issue functionality. Not sure yet if it's useful.

PR #467, opened 2025-01-17
Feature: Auxiliary scalars

### LINKED ISSUE #465: Feature: Transporting passive scalars
I'm working on adding in passive scalar transport, and I was wondering if anyone had any thoughts on whether the number of transported scalars should be specified at compile time or runtime. Some existing things are floating about to get started on this (e.g. m_nAux parameter, auxiliaries MultiFAB etc.), but there is also mention of implementing an NAUX variable in the index file.

ThomasHowarth:
This pull request may close #465 

Auxiliary variables are added through the following flags in the input: 
```
peleLM.aux_vars = a b c

peleLM.a.advect = 1 
peleLM.a.diff_coeff = 0

peleLM.b.advect = 0
peleLM.b.diff_coeff = 1e-2

peleLM.c.advect = 1
peleLM.c.diff_coeff = 1e-2
```
where the `advect` parameter determines whether or not the variable is transported, and the `diff_coeff` provides a diffusion coefficient.

The [test case](https://github.com/ITV-RWTH/PeleLMeX/tree/tlh_passive/Exec/RegTests/PassiveScalar) transports/diffuses a Gaussian profile through a periodic domain. For a 1m square domain, with 1m/s upward velocity, I see the following:
- $$t=0$$: 
image
- Scalar A (advected, not diffused):
   - $$t=0.5$$:
image
   - $$t=1$$:
image
- Scalar B (not advected, diffused):
    - $$t=0.5$$
image
    - $$t=1$$
image
- Scalar C (Advected and diffused):
    - $$t=0.5$$
image
    - $$t=1$$
image

# Discussion points:
1. User-defined function interfaces in the setup file (i.e. `pelelmex_initdata` and `bcnormal`) have been adjusted. I've done this to all cases, but when people pull this, their current cases may not compile. Not sure how to address this.
2. The implementation is done via the SDC iterations and is therefore not done in the case of `peleLM.incompressible = 1`
3. Currently the diffusion coefficient is specified as an input, rather than linked to a particular species/temperature. It should be relatively easy to link the coefficient to it since it is set in basically the [same place](https://github.com/ITV-RWTH/PeleLMeX/blob/3c423cbee05155d15e411ac7d83213c26259e07d/Source/PeleLMeX_TransportProp.cpp#L310), but I'm not sure what the choice should be.
4. Currently there is no source term implementation (e.g. for an age equation), however, when it is done, I suppose this could be integrated with @dmontgomeryNREL 's ODEQty in some fashion.
5. When filling the boundary conditions, the `s_aux` array passed to `bcnormal` needs to be [hardcoded](https://github.com/ITV-RWTH/PeleLMeX/blob/3c423cbee05155d15e411ac7d83213c26259e07d/Source/PeleLMeX_BCfill.H#L91). I've set this to `NVAR` and put a check to make sure that `m_nAux < NVAR`, but perhaps there is a more elegant way to do this.


baperry2:
Regarding your discussion points:

1. I don't like making changes that break things for users, but sometimes that can't be avoided. We should take a bit of extra time to make sure any interface changes are what we want and there aren't additional related changes we can do at the same time so that we do this as rarely as possible.

- I don't believe it's necessary to add `nAux` to `pelelmex_initdata`: instead the user could use `aux.nComp()` (an Array4 member function) to get this information as needed without modifying the interface.
- We could potentially avoid adding `s_aux` to `bcnormal` by stacking the aux data at the end of `s_ext` (which is just a pointer to AMReX::Real), but that's ugly and unintuitive, so a breaking change for the interface might be the lesser of two evils here.
- We have some other potential breaking changes in the pipeline to enable inflows on EBs (#156). It may make sense to delay merging this until those are done to minimize the number of times we break things for users

2. That's ok with me for now - essentially all real use cases for LMeX are not incompressible anyway. 
3. I think it makes sense to follow what we do when we have constant non-unity Le transport. Let users specify a Schmidt number (or Lewis/Prandtl) and define the scalar diffusion coefficients based on that Sc and the viscosity.
4. Yes, after merging this we will revisit the ODEQty stuff and fold it into this framework.
5. This is difficult because we need a copy of the Aux values so they don't get overwritten by calls to the bcnormal function that are applying BCs for other variables. We might not be able to avoid something akin to your current approach. But an improvement would be to create a new macro like `PELELMEX_MAX_NAUX` that can be set in the GNUmakefile and cmake commands rather than pinning the max to NVAR - allowing users to have more AUX variables for small mechanisms and not having so much unnecessary data allocated when NVAR is large. @marchdf - any ideas for a better solution when needing dynamically sized memory within a device function?

As always, I appreciate the new capability, which I'm sure will be quite useful to many.

ThomasHowarth:
I'll leave a note here of changes made/needed:
- [X] Remove nAux from `initdata`
- [x] Convert diff_coeff to Schmidt number
- [x] Figure out what to do with `s_aux`, both in terms of breaking `bcnormal` interface, and allocating its size.

 Additionally, I don't know why the Lint-codespell test has started failing - it seems to not like a naming convention for a variable that I have not touched.

baperry2:
A new version of codespell was released and it catches a few more things. The fix for that is in #469.

baperry2:
For the remaining issue of what to do about `s_aux`, I think we should be able to bypass creating the temporary array just to copy data back into the Array4 in the BCFill function. Instead of passing a pointer to amrex::Real, we could pass an `amrex::CellData` that gives direct access into the Array4 (see [Array4::cellData()](https://amrex-codes.github.io/amrex/doxygen/structamrex_1_1Array4.html)). The Aux values could also be populated with a separate `bcnormal_aux` function that only gets called when fillpatching the Aux variables, which would avoid the need mess with the existing `bcnormal` interface. @esclapez is working on a single struct that wraps all the case-specific IC/BC functions, which would make adding that additional function cleaner.

ThomasHowarth:
The CellData approach seems to work well. I think it gives a cleaner interface to the boundaries than the existing setup. It has also made the `m_nAux` input to the `bcnormal` function redundant, but we can remove that when an interface-breaking change comes. I also renamed this array to `aux_ext` since I think that makes more sense (I suppose the `s` in `s_ext` stands for state).

Source/PeleLMeX_BCfill.H
```diff
@@ -117,6 +116,79 @@ struct PeleLMCCFillExtDirState
...
+    // Get geometry data
+    const int* domlo = geom.Domain().loVect();
+    const int* domhi = geom.Domain().hiVect();
+    const amrex::Real* prob_lo = geom.ProbLo();
+    const amrex::Real* dx = geom.CellSize();
+    const amrex::Real x[AMREX_SPACEDIM] = {AMREX_D_DECL(
+      prob_lo[0] + (iv[0] + 0.5) * dx[0], prob_lo[1] + (iv[1] + 0.5) * dx[1],
+      prob_lo[2] + (iv[2] + 0.5) * dx[2])};
+
+    amrex::CellData<amrex::Real> aux_ext(aux.ptr(iv), aux.nstride, m_nAux);
```
baperry2: How about `auto aux_ext = aux.cellData(AMREX_D_DECL(iv[0], iv[1], iv[2]));` instead here?

baperry2 (changes requested):
To avoid having to define the `bcnormal_aux` function all over the place, let's add the changes in https://github.com/ITV-RWTH/PeleLMeX/pull/3 (open to other ideas if you have a cleaner way to do that)

Can you set this up to run in the testing suite? See https://github.com/AMReX-Combustion/PeleLMeX/pull/441 for an example - but this doesn't need the additional python script validation check

ThomasHowarth:
Has the PR broken itself? Seems to be stuck checking something

Source/PeleLMeX_BCfill.H
ThomasHowarth: The cellData member function needs three arguments, so I've had to explicitly ifdef this call this way. The cellData member function may need an IntVect overload to avoid this in the future.

ThomasHowarth:
> To avoid having to define the `bcnormal_aux` function all over the place, let's add the changes in [ITV-RWTH#3](https://github.com/ITV-RWTH/PeleLMeX/pull/3) (open to other ideas if you have a cleaner way to do that)

For this, I've taken a slightly different approach and wrapped the function call in the ifdef, with an abort if someone tries to run a case with Dirichlet boundaries with auxiliaries without defining a bcnormal function - does this seem reasonable, or would a template function be better to do you think?

> Can you set this up to run in the testing suite? See #441 for an example - but this doesn't need the additional python script validation check

- [x] Tests

Source/PeleLMeX_BCfill.H
baperry2: Ah yeah, I figured there was an overload in 2D that just used i,j but because it's indexing into an Array4 it makes sense that it expects i,j,k even when compiled in 2D.

Adding an IntVect version of cellData would be a simple PR into AMReX if you're interested in doing that.

baperry2 (approved):
Good to go once the GPU compilation issues are fixed. I think you just need to replace `std::hypot` with `std::sqrt(a**2 + b**2)` because it looks like the former isn't supported with HIP.

baperry2:
Actually @ThomasHowarth have you tested this for EB? I did something quick and it looks like scalars are diffusing properly with EB but they are not advecting when EB is turned on.

ThomasHowarth:
> Actually @ThomasHowarth have you tested this for EB? I did something quick and it looks like scalars are diffusing properly with EB but they are not advecting when EB is turned on.

I haven't tested with EB, but I thought I'd caught all of the parts of the code with it. I'll make a test and check.

ThomasHowarth:
@baperry2 I'll fix the conflicts, but I just tried with the FlowPastCylinder case, and the scalars worked fine - could you be more specific about a case that didn't work?

baperry2:
If I compile the 2D PassiveScalar case with `USE_EB=TRUE` and then run with the additional flag `eb2.geom_type=all_regular`, which simply runs without any EB and should give the same results, then I get NaNs for the passive quantities starting in the second plt file generated. 

@bssoriano noticed some issues when corse/fine boundaries intersect with inflows. Perhaps this is related.

ThomasHowarth:
I've traced the issue to the auxiliary boundaries not being filled properly at initialisation, but I can't figure out where the problem is. If you compile in debug mode and turn on the FPE catches, it crashes when looking for some boundary data (regardless of with/without EB). I've probably missed a fillpatch somewhere, but I can't spot it—I'll keep working on it.

Source/PeleLMeX_Data.cpp
```diff
@@ -52,6 +52,7 @@ PeleLM::LevelData::LevelData(
   }
   if (a_nAux > 0) {
     auxiliaries.define(ba, dm, a_nAux, a_nGrowState, MFInfo(), factory);
+    diff_aux_cc.define(ba, dm, a_nAux, a_nGrowState, MFInfo(), factory);
```
baperry2: Maybe it's a number of ghost cell issue? Shouldn't `diff_aux_cc` have the same number of ghost cells as `diff_cc`?

ThomasHowarth: I've fixed this, but there were too many ghost cells rather than too few, so I don't think this was an issue.

ThomasHowarth:
Thanks to some work from @bssoriano, we spotted the missing boundary data. I've also made some things more strongly typed (more const's and constexpr's) and gone through and used the m2c/c2m unit conversions in the kernel functions where possible.

baperry2:
Some final fixes that let this run with EB in debug mode: https://github.com/ITV-RWTH/PeleLMeX/pull/5

A last thing is that it looks like we don't apply redistribution for the Aux variables with EB, but that can wait. I'd like to get this merged before #477.

baperry2:
Thanks again @ThomasHowarth for the new capability and @bssoriano for helping to find the fillpatching issues. I think this is in good enough shape to merge now, although some of the EB stuff might need to be revisited eventually.

PR #468, opened 2025-01-24
Add option to plot external sources

d-montgomery:
This PR includes a runtime option `peleLM.plot_extSource` for adding external sources to the plot files.


baperry2 (approved):
Once #469 merges pull that in to fix the codespell issue and this is good to go.

PR #470, opened 2025-01-25
From prior commit: fix bug with indexing in plt when LES is not active 

baperry2:
In the plotting routine, the counter should only increment for plotting the LES stuff when using LES.

@dmontgomeryNREL: note that this means all the source terms you've been looking at are probably off by one - it may not actually be the enthalpy source term that's going crazy in your case.


d-montgomery (approved):
Thanks for catching this!

PR #472, opened 2025-01-29
Template BCFill functions to reduce repeated code

baperry2:
Leaving as a draft for now pending further testing and other PRs currently in the queue that affect the same area of code. 

May also be able to do something similar to eliminate some of the `PeleLM::fillpatch_XXX` functions in PeleLMeX_BC.cpp.


ThomasHowarth:
I think the same thing could be done for the `PeleLM::getXXXVect/averageDownXXX` functions also

PR #477, opened 2025-02-09
Add InitialBoundaryCondition (IBC) struct.

esclapez:
Same principle as in [PeleC](https://github.com/AMReX-Combustion/PeleC/pull/743) but for kernels only at this point:
the initial and boundary condition kernels are gathered in a struct inheriting from a default, such that the user only has to override the ones he really need. It offers more flexibility to introduce change without breaking backward compatibility in the future.

Only a few Exec updated for now and starting with a draft PR, let's discuss any (other) breaking changes in those kernels @baperry2 .
Following the discussion in #467, I've removed `nAux` from `initdata` and added `s_in` which can be used to do some same-face inflow/outflow tricks.

We can probably also move some other functions in there (EBGeom, UserDefinedDerived, ExtSourceTerm, ...) although I'm not sure what get copied to the device when a static kernel from the struct is called within an AMReX macro+lambda.


esclapez:
- [x] Update Exec/Regtests
- [x] Update Exec/Plasma
- [x] Update Exec/Production
- [x] Update the documentation/tutorials
- [ ] Add warning and how to move from old to new setup

Source/PeleLMeX_TransportProp.cpp
```diff
@@ -469,7 +470,7 @@ PeleLM::getDiffusivity(
         const auto& diff_ec = beta_ec[idim].array(mfi);
         amrex::ParallelFor(
           ebx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
-            zero_visc(
+            lprobIBC.zero_visc(
```
baperry2: Call as `ProbIBC::zero_visc` to avoid instantiating a ProbIBC object?

Source/PeleLMeX_FlowController.cpp
```diff
@@ -77,14 +78,17 @@ PeleLM::initActiveControl()
...
       dumbx,
-      [x, nAux = m_nAux, s_ext_d, ctrl_flameDir_l, time_l, geomdata, lprobparm,
-       lpmfdata] AMREX_GPU_DEVICE(int /*i*/, int /*j*/, int /*k*/) noexcept {
-        bcnormal(
-          x, nAux, s_ext_d, ctrl_flameDir_l, 1, time_l, geomdata, *lprobparm,
+      [fake_state, x, s_ext_d, ctrl_flameDir_l, time_l, geomdata, lprobparm,
+       lpmfdata,
+       lprobIBC] AMREX_GPU_DEVICE(int /*i*/, int /*j*/, int /*k*/) noexcept {
+        const auto s_in = fake_state.cellData(0, 0, 0);
+        lprobIBC.bcnormal(
```
baperry2: Call as ProbIBC::bcnormal to avoid instantiating the object?

Source/PeleLMeX_TransportProp.cpp
baperry2: Also, while we're mucking around with this interface, how about adding probparm to the zero_visc interface? I can imagine potential uses where users might want to be able to change where zero_visc gets applied at runtime.

Source/PeleLMeX_DefaultIBC.H
```diff
@@ -0,0 +1,178 @@
...
+    const amrex::CellData<amrex::Real>& /*s_in*/,
+    amrex::Real* /*s_ext[NVAR]*/,
+    const int /*idir*/,
+    const int /*sgn*/,
+    const amrex::Real /*time*/,
+    amrex::GeometryData const& /*geomdata*/,
+    ProbParmDefault const& /*prob_parm*/,
+    pele::physics::PMF::PmfData::DataContainer const* /*pmf_data*/)
+  {
+    /** Notes:
```
baperry2: Add an `amrex::Abort()` here so if the function gets called without being overridden it doesn't cause unexpected behavior for the user

Source/PeleLMeX_DefaultIBC.H
```diff
@@ -0,0 +1,178 @@
...
+    int /*j*/,
+    int /*k*/,
+    int /*is_incompressible*/,
+    amrex::Array4<amrex::Real> const& /*state*/,
+    amrex::Array4<amrex::Real> const& /*aux*/,
+    amrex::GeometryData const& /*geomdata*/,
+    ProbParmDefault const& /*prob_parm*/,
+    pele::physics::PMF::PmfData::DataContainer const* /*pmf_data*/)
+  {
+    /** Notes:
```
baperry2: Add an amrex::Abort() here so if the function gets called without being overridden it doesn't cause unexpected behavior for the user

Source/PeleLMeX_EBUserDefined.H
```diff
@@ -25,39 +25,5 @@ EBUserDefined(
   Abort("In default EBUserDefined function! Shouldn't be here. Copy and edit "
```
baperry2: Why not also include `EBUserDefined` in ProbIBC? That way everything case-specific would be in the same place, making things simpler if we move toward compiling multiple cases at the same time.

Source/PeleLMeX_BCfill.H
```diff
@@ -8,6 +8,7 @@
 
 using namespace amrex;
 
+template <typename T>
```
baperry2, line 11: I don't think templating here is strictly necessary, as we always use `ProbIBC`, but it's fine to leave it because if we do move to compiling multiple cases at once then we would need the templating.

Source/PeleLMeX_BCfill.H
```diff
@@ -77,7 +87,8 @@ struct PeleLMCCFillExtDirState
           }
           // bcnormal() is defined in pelelmex_prob.H in problem directory in
           // /Exec
-          bcnormal(x, m_nAux, s_ext, idir, 1, time, geom, *lprobparm, lpmfdata);
+          T::bcnormal(
+            x, s_in, s_ext, idir, 1, time, geom, *lprobparm, lpmfdata);
```
baperry2, line 90: I think the CodeQL warning here is spurious due to NUM_SPECIES varying from case to case

baperry2 (commented):
Thanks @esclapez - this is awesome! Looks pretty close to merging.

I left some specific comments, but I also have a few general thoughts:

1. Although the new struct is named `ProbIBC`, I imagine we will quickly want to include problem-specific stuff that isn't strictly initial/boundary conditions (e.g, external source terms, as currently defined in PeleLMeX_ProblemSpecificFunctions.cpp). What would you think about calling the new struct `ProblemSpecificFunctions`, matching PeleC? Or is it not your intention that that stuff like external source terms would become part of this struct? I'm also ok with leaving as `ProbIBC` but still putting non-IBC stuff in the struct.
2. After this, we should be able to merge the EB-Inflow capability without further breaking changes to existing cases, right? Let's try to get any major breaking changes done right now to minimize disruption to users.
3. Things get a bit messy with probparm when it requires dynamically sized data (I had to add a temporary solution to freeing memory for these in #480). I think we can do something better for that that will also involve breaking changes for case files - I'll try to set that up so it can be merged in a separate PR at the same time as this.

baperry2:
Actually what I was going to do for (3) above gets pretty messy. I think the best solution might be to add a null `PeleLM::freeProbParm()` function for all cases as part of this PR.

Source/PeleLMeX_BCfill.H
esclapez: That's my guess too, it might not have been a problem before bcnormal got static.

Source/PeleLMeX_BCfill.H
esclapez: My thought exactly.

esclapez:
@baperry2 Thanks for the feedback !
I've changed things around to follow your suggestions in 1). The closer to PeleC the better, even though ProblemSpecifcFunctions is a mouthful (or a keyful!) :)
For 2), yes adding the inflow EB should not change further the signature of setEBType and setEBState, but if you think of something we're missing in there we might as well add it now.
For 3) and your later comment, I'm not sure how you want to handle it. Do you want to add a function to the ProblemSpecificFunctions struct for deallocating data in ProbParm and let the user handle it, or directly alongside the readProbParm in the .cpp ? Feel free to push directly into the PR branch.

drummerdoc:
Candace Gilet was recently asking about mixing Neumann and Dirichlet conditions on the EB... Without digging at all into the new structure of all this, I don't know much of what I'm talking about, but is there a mechanism for carrying the transport coefficients together with the BC's so that the use can selectively zero them if appropriate (like zeroVisc)?

baperry2:
@drummerdoc: This exists in a form for setting isothermal/adiabatic EBs, where the `setEBType` function acts like `zero_visc` at the EB: https://amrex-combustion.github.io/PeleLMeX/manual/html/Tutorials_BFSFlame.html#problem-specifications

I don't think anything similar is set up for other state quantities, but I'd imagine the same strategy could be applied for those pretty easily as well. @esclapez can correct me if I'm wrong.

@esclapez - I'll push the changes for `freeProbParm()` next week. Other than that I think this should be good to go :)

drummerdoc:
Nice! thanks for the pointers, Bruce.

esclapez:
@drummerdoc
In the implementation of EBInflow right now, you flag your EB cells with one of "WallNoSlipAdiab", "WallNoSlipIsoT" or "Inflow". I think adding "WallSlipAdiab", "WallSlipIsoT" should be relatively easy since we have the EB normal and with the velocity in the first cell we should be able to add a simple wall model. Default is "WallNoSlipAdiab". That should cover most basis and we should be able to add anything custom too.

baperry2:
@esclapez - I created a pull request into this branch with the changes for freeProbParm.

After you approve that, let's create a list of changes that users will need to make as a result of this PR, then merge this!

esclapez:
@baperry2 I'm adding a tweak to address Marc's comment. I actually removed the ability to have partial coverage of the EB face with a EBtype different from the default in the EBInflow implementation. With this tweak, `setEBType` will return an `int` for the BC type and a `real` to indicate the fraction of the EB face covered by that type. It will be used to scale the diffusion coeffs.

baperry2:
As we get ready to merge, @marchdf and @dmontgomeryNREL suggest that we create a numbered release before and after to help users manage the breaking changes. Before you had been doing releases numbers as YEAR.MONTH (like AMReX), but we haven't released anything since 23.11. We could also move to the standard MAJOR.MINOR.PATCH numbering scheme (like [AMR-Wind](https://github.com/Exawind/amr-wind?tab=readme-ov-file#versioning-and-tags)). 

@esclapez and @drummerdoc: any thoughts on this and how else to help people manage versioning especially with breaking changes as we gain more users?

drummerdoc:
Good idea with the versioning.  I'm ok using the month.year scheme and there just being missing months...maybe add a note to any version that skips months/years just to be clear.

esclapez:
I was thinking of indeed releasing a new version because it's been a while. The YY.MM format makes sense with frequent releases but there hasn't been a strong motivation to do so with development slowing down after ECP. So maybe switching to MAJOR.MINOR.PATCH would make more sense.

I was trying to find a way to detect at compile time if the user still has a `pelelmex_prob_parm.H` and point to a page in the doc with the necessary changes.

baperry2:
How about releasing one last release in the MM.YY format before this merge, then releasing this version as `1.0.0`. I think I prefer MAJOR.MINOR.PATCH going forward as it's more informative about what has actually changed. 

Autodetecting and pointing to documentation when `pelelmex_prob_parm.H` is present would be awesome!

baperry2 (approved):
@esclapez - I implemented your suggestion of raising an error if the `pelelmex_prob_parm.H` file is present and also added step by step instructions for users to follow to port their cases for these changes.

baperry2:
@esclapez, @drummerdoc, and @ThomasHowarth: I will merge this later this week and release the versions as described above unless anyone objects.

PR #481, opened 2025-02-13
Add check during ODEQty initialization

d-montgomery:
Ensure user defined ode names are set properly.


Source/PeleLMeX_Setup.cpp
```diff
@@ -910,6 +910,10 @@ PeleLM::variablesSetup()
     Print() << " First ODE: " << FIRSTODE << "\n";
     set_ode_names(m_ode_names);
```
baperry2, line 911: also check that `m_ode_names.size() == NUM_ODE`

baperry2 (changes requested):

baperry2 (approved):

PR #482, opened 2025-02-14
Add option to plot state before init proj

baperry2:
PeleLMeX doesn't dump a plot file until after initial projections/iterations. When setting up new cases, this can make debugging hard, because if there are problems in the initdata function, the solver will crash before dumping a plot file, so the problems aren't easily visualized. 

This PR adds an option `amr.plot_init_state` that, when activated, dumps a plot file after initdata but before any of the initial projection/iteration stuff. The plot file gets named `plt000-1`.


esclapez:
Note that you can also bypass the initial projection and iterations such that `plt00000` contains initdata:
```
peleLM.num_init_iter = 0
peleLM.num_divu_iter = 0
peleLM.do_init_proj = 0
```

baperry2:
Yep this just makes things a bit easier and lets you get the initial condition without deactivating those.

d-montgomery (approved):
I like this option. When using ParaView to visualize, the `-1` makes it so the initial state must be opened separately.  This is probably a good thing as it keeps the initial state separate from the state after the initial projections.

PR #483, opened 2025-02-14
change macro USE_EFIELD to USE_PLASMA

EnnaDelfen:
In this PR I'm changing the pele macro USE_EFIELD to USE_PLASMA to extend the plasma capabilities beyond efield. 
At this point it's only a matter of semantics but more PR with additional model options will follow.


baperry2:
If you and @esclapez think that PLASMA is a better name than EFIELD, I'm fine with this change.

For consistency though, should some other things be renamed? For example `NUMEFIELD` and the `Efield` directories in `Source/` and `Exec/`, and any references to EFIELD in the docs. I guess exactly what gets renamed would depend on what additional plasma capabilities will be added. What do you have in mind?

Don't worry about the pending test, that's just because github expects required tests and the name of the required test was changed. I'll fix that after we merge this.

EnnaDelfen:
@esclapez suggested we do the updating of the Source and Exec folders in a second PR, but if it's OK with the both of you I'll add it to this one. Good point about the references to EFIELD in the docs, I'll also take care of this.
We want to add more generic plasma capabilities, for instance plasma generated by MW, which presents no charge separation and thus doesn't require the whole NL solve.

esclapez (approved):

PR #492, opened 2025-03-06
Use ubuntu 24 in all CI

jrood-nrel:


baperry2:
Does this resolve the issues we had with MPI and Ubuntu 24 (discussion in #443)?:

jrood-nrel:
Yeah it does. MPICH in ubuntu 24 is broken so just using OpenMPI for now instead. I also am outputting the ctest runs to see that MPI is working as expected.

baperry2 (approved):

PR #494, opened 2025-03-06
move reset covered mask outside if to avoid undefined usage

### LINKED ISSUE #491: m_resetCoveredMask may not be initialized
If I'm not mistaken, the m_resetCoveredMask variable may not be initialized in certain cases. It is initialized in PeleLMeX_Init.cpp line  72 if (max_level > 0) and line 321  if (!m_restart_chkfile.empty()). If max_level == 0 and m_restart_chkfile.empty() the value of m_resetCoveredMask is undefined when going through the check if (m_resetCoveredMask != 0) in PeleLMeX_Utils.cpp line 696. This leads to random behavior in PeleLM::resetCoveredMask().

baperry2:
Closes #491 

Something similar was done in 55c3e67 for regridding, but it looks like this was missed, which can lead to entering resetCoveredMask with `m_resetCoveredMask` in an undefined state when initializing single level simulations as found by @QuentinMale

### DIFF
--- a/Source/PeleLMeX.H
+++ b/Source/PeleLMeX.H
@@ -1675,7 +1675,7 @@ public:
 
   // Fine-covered cells mask
   amrex::Vector<std::unique_ptr<amrex::iMultiFab>> m_coveredMask;
-  int m_resetCoveredMask;
+  int m_resetCoveredMask{0};
 
   // Chemistry BA & DM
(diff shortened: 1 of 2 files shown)


esclapez (approved):
That should work, but I would also initialize that variable to 0 in PeleLMeX.H. I'm not sure why it's not already the case.

PR #499, opened 2025-04-02
pass prob parm as pointer to user defined source term function

baperry2:
Fixes issue when running on GPU


d-montgomery (approved):
Thanks for trouble shooting this. I tested it this morning on GPU and everything works now.

PR #501, opened 2025-04-09
Workarounds for bcnormal being called for walls

baperry2:
bcnormal is called for any boundary that is [labeled as ext_dir](https://github.com/AMReX-Combustion/PeleLMeX/blob/8ae6a1f9817f202adb46e77e94c81971d8641804/Source/PeleLMeX_BCfill.H#L64), which includes [both inflow and NoSlipWalls](https://github.com/AMReX-Combustion/PeleLMeX/blob/8ae6a1f9817f202adb46e77e94c81971d8641804/Source/PeleLMeX_BC.cpp#L18). This is non-intuitive behavior which could cause confusion for users. 

Add a note about this behavior in the documentation. 

Remove an input file which results in confusing behavior because of this.


baperry2:
@malihass: FYI

d-montgomery (approved):

PR #502, opened 2025-04-13
Extract velocity field from an outflow plane

drummerdoc:
Adds code in PeleLMeX::Advance to optionally write a FAB slice file of the velocity field using level 0 data from a specified plane of the simulation (and appends the current time for later use).  Can be turned on during any restart, and provides options for setting the frequency (in time steps), and folder and filename prefixes.  Note that this will overwrite any identically-named files that existed previously.

Code to assemble these slices into a "turbulence" file used by the turbInflow code was added to AMReX in this  [pull request](https://github.com/AMReX-Codes/amrex/pull/4418),, and the turbInflow code modifications required to read these assembled files is added in a [companion](https://github.com/AMReX-Combustion/PelePhysics/pull/566) for PelePhysics.  I've done some cursory testing of this capability, but more extensive testing should be done and the documentation should be modified to explain how it works and how it is to be used.

### DIFF
--- a/Source/PeleLMeX_Advance.cpp
+++ b/Source/PeleLMeX_Advance.cpp
@@ -243,11 +243,6 @@ PeleLM::Advance(int is_initIter)
   }
 
   //----------------------------------------------------------------
-
-  // Manage outflow data
-  WriteOutflowPlane(m_nstep, m_cur_time);
-
-  //----------------------------------------------------------------
   // Wrapup advance
   // Timing current time step
   if (m_verbose > 0) {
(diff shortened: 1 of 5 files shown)


Source/PeleLMeX_Advance.cpp
```diff
@@ -253,6 +258,47 @@ PeleLM::Advance(int is_initIter)
...
+  if (
+    m_write_outflow_plane >= 0 && m_write_outflow_plane < 2 * AMREX_SPACEDIM &&
+    a_step >= 0 && a_step % m_write_outflow_plane_int == 0) {
+    const int outflow_dir = m_write_outflow_plane % AMREX_SPACEDIM;
+    const Orientation::Side outflow_face =
+      m_write_outflow_plane >= AMREX_SPACEDIM ? Orientation::high
+                                              : Orientation::low;
+    const Orientation orient(outflow_dir, outflow_face);
+
+    Box outflow_box = amrex::adjCell(geom[0].Domain(), orient, 1);
```
baperry2: How about generalizing this to dump from intermediate planes as well? If we dump data directly on boundaries the data may be affected somewhat by outflow BCs (if non-periodic).

drummerdoc: Sure, but then we'd have to rename everything! :). My vote is to leave as is and generalize when the need arises, only because it's slightly harder to make sure it's right, and complicates the specification a bit.

Unrelated followup though...in my tests, I chopped data from EB_PipeFlow, but shoved it into a non-EB case.  It "worked" but gave some very odd looking results that might have just reflected the lack of care in ensuring a physical union of the two runs (properly dealing with inflow+transport, etc).

drummerdoc: or.... could do `plane_dir`, `plane_loc`.  IF that works, I could fix things.

baperry2: Yeah that's what I was thinking. Just have `plane_loc` be an int that indicates the coarse grid index of the plane being dumped. Or a real and then compute the int index.

drummerdoc: Ok, should be ready

baperry2:
There appears to be an issue with the data output by this for y and z planes (x planes work fine). But I think instead of fixing that, I'll modify the script that converts the planes to turbinflow format to instead read the planes that we dump using DiagFramePlane, which would avoid the need to add this new capability to PeleLMeX.

PR #511, opened 2025-05-15
Diagnostic doc move, dump on 0th iter

baperry2:
Move most of the documentation to PelePhysics (https://github.com/AMReX-Combustion/PelePhysics/pull/572) because it is shared across the Pele codes.

Dump diagnostics on initialization from scratch (but not from a chk file, because the diagnostic should have already been saved in the previous run)


malihass (approved):
Just tested it in a turbulent channel flow case, and it works great, thanks for that!

PR #512, opened 2025-05-16
Fix memory leak for Soret isothermal boundaries

baperry2:
Some multifabs for setting up Soret isothermal BCs are allocated with `new` but never `delete`d. This causes a memory leak that was noticed by @bssoriano.

### DIFF
--- a/Source/PeleLMeX_Diffusion.cpp
+++ b/Source/PeleLMeX_Diffusion.cpp
@@ -453,6 +453,11 @@ PeleLM::correctIsothermalBoundary(
     }
   }
   // TODO: wbar fluxes disabled for this case - boundary system becomes complex
+  for (int lev = 0; lev <= finest_level; lev++) {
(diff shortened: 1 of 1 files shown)


ThomasHowarth (approved):
My bad - thanks for this Bruce; looks good.

Just as a note on the Soret stuff, we're working on making a NUM_LITE_SPECIES in the mechanisms and reducing the size of the coefficient MultiFAB which should reduce memory usage, particularly for large mechanisms.

ThomasHowarth:
I'm happy with how it is currently, but we could also use a ``std::unique_ptr<MultiFab>`` instead of ``new`` and the memory cleanup should be handled for us, right?

baperry2:
There's probably a nicer way to do this with `std::unique_ptr`, which is what I tried first, but that led to type conversion issues for the vector of arrays of multifab pointers, so I just defaulted to adding `delete` rather than messing around with function interfaces.

Sounds good on moving `NUM_LITE_SPECIES`.

PR #515, opened 2025-05-18
EB-Inflow

### LINKED ISSUE #473: Enable inflow on EBs

baperry2:
Import EB-Inflow capability from @esclapez for new changes based on #477 (v1.0.0)

TODO:
- [x] Port test case with an EB-inflow to PeleLMeX v1.0.0 prob file formatting
- [x] Doc updates

@escalpez - let me know if you've already done this in a branch somewhere!


baperry2:
Closes #473

jrood-nrel (approved):

esclapez:
@baperry2 Sorry, I did have that somewhere but I've got a bit too much on my plate lately.
We can move forward with the PR to have this feature available in dev, but a couple of things to keep in mind:
 - currently no scalar diffusion on eb-inflow
 - I'm not sure the SFINAE code in EB_BCfill.H is still useful with the PSF struct.
 - keeping separate EB_inflow and EB_isoT is pretty limited. I was thinking of having a params `EB2.bc_type` that defaults to `wall_no_slip_adiab`, but can be changed to `wall_no_slip_isoT` (+ parsing a single value for T) or `wall_slip_adiab` if/when we have wall models. Or `EB2.bc_type = UserDefined` which trigger the use of the `bcnormal_eb`, `bctype_eb`.
 
Maybe those things can be adressed down the line (hopefully not too far), but I would add warning on that capability in the doc for now saying that it's still under development.

baperry2:
Thanks for the comments @esclapez. We do need this capability in dev for a current project, so I'll probably merge with the code as is and we can add more refinement later.

- Scalar diffusion on EB-inflow would be good to have, but most of the time advection probably dominates immediately at the inflow face so I'm fine with having it as is for the moment.
- The SFINAE stuff isn't strictly necessary but it does fail a bit more gracefully so I left it in. 
- I will add some warnings in the docs and perhaps in the code that this capability is still in development and may change.

esclapez (approved):

PR #517, opened 2025-05-28
Update/extend TurbInflow Test Cases for Improved Turbinflow Capability

baperry2:
Add new tests that validate the improved TurbInflow capability from https://github.com/AMReX-Combustion/PelePhysics/pull/577

The fully periodic convected vortex case is used to create inflows in two ways: by saving outflow planes with the DiagFramePlane capability and assembling into a temporally dependent inflow, and by converting a single 3D plt file into a turbinflow that can be moved through in the normal direction using Taylor's hypothesis. Both are tested by converting the y-direction to inflow/outflow boundaries and applying the turbinflows on the ylo boundary.

The results of the two approaches, which can be generated using the test case being added in this PR:

**Inflow from Planes:**
x_velocity_output
mag_vort_output

**Inflow from normal dir of 3D plt:**
x_velocity_output
mag_vort_output

Will also need to clean up the legacy TurbInflow tests as part of this PR.


drummerdoc (approved):
Looks beautiful!  Can't test in detail tho.

malihass:
Super good work! And I prefer the diagframe approach which also allows to visualize the plane we extract (not easily doable with the previous approach)

One surprising thing to me is that output of the test that reads the diagframe seems to do an interpolation over z, rather than over time. Are you still interpolating input frames based on the timestamp? See below

```
 ====================   NEW TIME STEP   ==================== 
 STEP [118] - Time: 0.005141735297, dt 5e-05
   SDC iter [1] 
Reading new data because z 0.2595867649 is outside 0.2265625 and 0.2578125
read_turb_planes filling 15 to 18
 --> now have interp data for range [0.2421875, 0.2734375) with current z = 0.2595867649
 ```
 
 
 EDIT: Looking at the pelephysics documentation, you added this
"The behavior of the utility depends on the periodicity of the data. If the data is periodic in the normal direction, it is treated as spatial data, the normal direction is traversed based on the specified turb_conv_vel, and the inflow data can be recycled to allow arbitrarily long simulations": So is it because the precursor sim is periodic?

baperry2:
The turbinflow utility just always refers to its 3rd dimension as `z` because it was initially set up for spatial data. With temporal data it does interpolate based on timestamp. From a practical perspective it doesn't it doesn't matter much, except that it doesn't recycle back to the start if the third dimension is time. I think I can change the print statements to make this less confusing.

malihass (approved):
Took me a while to reproduce what I wanted to see, but it is working perfectly
I also tested cases where we query the inflow out of bounds, and it is exiting as expected
I havent tested the non periodic Turbulent channel case so far, but we can do this later.
Congrats!

PR #521, opened 2025-06-02
Single droplet evaporation validation

d-montgomery:
This PR migrates and updates the original [PeleMP validation tests](https://github.com/AMReX-Combustion/PeleProduction/tree/Landon/soot_and_spray/PeleMPruns/single_drop_tests) for single droplet evaporation.  It includes:

- Additional validation cases from Nomura et al., Wong & Lin, and Runge et al.
- Updated input parameters, validation data, and figure generation
- Documentation included in PelePhysics [PR #581](https://github.com/AMReX-Combustion/PelePhysics/pull/581)

I will be working on adding new features in PelePhysics for calculating liquid fuel properties, which should help the results from the Nomura et al. case.

nomura_res_2025
wonglin_res_2025
daif_res_2025
runge_simple_res_2025


Exec/RegTests/SprayEvap/ref_files/RungeJP8/refdvals.csv
```diff
@@ -0,0 +1,18 @@
+X,Y
```
baperry2: `t, dd02` rather than `X,Y`?

Exec/RegTests/SprayEvap/ref_files/Daif/refTvals.csv
```diff
@@ -0,0 +1,18 @@
+t,T,min_uncertainty,max_uncertainty
```
baperry2: Not a huge fan of including data files like this, but these are all mall and necessary for validation so its ok here.

Exec/RegTests/SprayEvap/pelelmex_prob.cpp
```diff
@@ -0,0 +1,77 @@
...
+  amrex::Real massfrac[NUM_SPECIES] = {0.0};
+  massfrac[N2_ID] = PeleLM::prob_parm->Y_N2;
+  massfrac[O2_ID] = PeleLM::prob_parm->Y_O2;
+  amrex::Real T_g = PeleLM::prob_parm->T0_gas;
+  amrex::Real T_eff = (2. * T_d + T_g) / 3.;
+  amrex::Real p_cgs = m2c::P(PeleLM::prob_parm->P_mean);
+  amrex::Real rho_cgs = 0.;
+  eos.PYT2R(p_cgs, massfrac, T_eff, rho_cgs);
+
+  bool FA = false;
```
baperry2: These seem unnecessary to me and make the code less readable IMO. Probably the most readable is to make a separate variable for each like `get_mu`, `get_chi`, etc. They can also all be `const`.

Exec/RegTests/SprayEvap/GNUmakefile
```diff
@@ -0,0 +1,39 @@
...
+FSANITIZER = FALSE
+THREAD_SANITIZER = FALSE
+
+# PeleLMeX
+USE_PLASMA = FALSE
+USE_PARTICLES = TRUE
+SPRAY_FUEL_NUM = 3
+
+# PelePhysics
+Chemistry_Model = fuels_nonreacting
```
baperry2: For this to work you'll have to update the PelePhysics submodule to capture your changes there.

Exec/RegTests/SprayEvap/Validate.py
```diff
@@ -0,0 +1,201 @@
...
+| WongLin()    | decane         |
+| Daif()       | heptane/decane |
+| RungeHep()   | heptane        |
+| RungeDec()   | decane         |
+| RungeMix()   | heptane/decane |
+| RungeJP8()   | POSF10264      |
+"""
+
+# Case object
+case = WongLin()
```
baperry2: How about allowing the case to be a command line input?

Exec/RegTests/SprayEvap/Validate.py
```diff
@@ -0,0 +1,201 @@
...
+            f"rm -rf {case.name}/plt* {case.name}/*.p3d {case.name}/pele_vals.csv"
+        )
+
+    # Create case-specific input file
+    CreateInputFile(case)
+
+    # Get the Pele executable
+    exe = ""
+    for f in os.listdir(FILE_PATH):
+        if f.startswith("Pele") and f.endswith(".ex"):
```
baperry2: also ensure the executable contains `MPI`. Perhaps raise an error if multiple executables are found.

Exec/RegTests/SprayEvap/Validate.py
```diff
@@ -0,0 +1,201 @@
...
+            exe = f
+    if not os.path.exists(exe):
+        error = "Pele executable not found"
+        raise ValueError(error)
+
+    # Run the case
+    os.system(f"mpiexec -np 6 ./{exe} {case.input_file}")
```
baperry2: Maybe let the number of the processors be an optional command line argument. 6 processors is probably too many if this is run as part of the CI.

baperry2 (changes requested):
How about calling this case `SingleDropEvap` to be more clear?

To ensure this doesn't break, it should run in the CI. Probably the best way to do this is to add the base case to the CMake tests in `ci.yml`. That would leave the python part untested but ensure the Pele part keeps working. To run the python stuff you'd have to choose one of the faster cases and run that in `linux.yml`.

I get a `deltaT iters not converged` error for the WongLin case. Didn't run everything to completion but others seem to be working.

Exec/RegTests/SprayEvap/Validate.py
d-montgomery: I'll add the number of processors as an optional argument and will set up the CI to run the WongLin test for a limited number of time steps via the input_general.inp file instead of using Validate.py.

d-montgomery:
- [x] Fix headers in .csv's
- [x]  Clean up transport function call in prob.cpp
- [x] Rename mechanism in GNUmakefile
- [ ] Point to updated PelePhysics with liquid_fuels_nonreacting mechanism
- [x]  Check if exe includes MPI in Validate.py
- [x] Add to CI

d-montgomery:
@baperry2, the CI test for SingleDropEvap is failing due to an unused variable, `mu_parts` in `SpraySB.cpp` ([see line 53](https://github.com/AMReX-Combustion/PelePhysics/blob/fc3eae4e9a121707b084324c77b491567f63a712/Source/Spray/SpraySB.cpp#L53)).  This wasn't triggered previously as the other spray tests have `SPRAY_FUEL_NUM = 1`.

It does not appear that `mu_parts` is useful in SpraySB.cpp.  Should I delete it entirely or use an `amrex::ignore_unused` for now?

baperry2:
Yeah it's better to just get rid of the unused variables if they are truly unused. `amrex::ignore_unused()` is more for function arguments and that type of thing where you cant just comment/delete the offending unused variables.

d-montgomery:
Everything seems to be working now. The last step is to merge PelePhysics [#581](https://github.com/AMReX-Combustion/PelePhysics/pull/581) and update the corresponding submodule in this LMeX PR.

baperry2 (approved):

PR #527, opened 2025-06-05
Avoid set but not used variable on GPU

jrood-nrel:


Source/PeleLMeX_K.H
```diff
@@ -1190,13 +1190,11 @@ check_divu_dt(
                              : std::abs(rho(i, j, k)) / denom;
     }
   }
-#ifndef AMREX_USE_GPU
   if (a_dt > dtcell) {
     amrex::Abort("Error: divU dtcell larger than actual dt !");
   } else if (a_dt > dtcell * dtfactor) {
     amrex::Print() << "Warning: divU dtcell close to actual dt ! \n";
```
baperry2: Let's change this to an `amrex::Warning()` rather than `amrex::Print()`, then we shouldn't need any `#ifndef AMREX_USE_GPU`. Even on CPU it should be Warning rather than Print because this could be happening on any rank.

baperry2 (approved):

PR #545, opened 2025-07-30
update PelePhysics

baperry2:
Expect some minor diffs in unit tests due to https://github.com/AMReX-Combustion/PelePhysics/pull/595 but maybe it's causing some tests to fail?

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -129,7 +129,7 @@ jobs:
     - name: Run
       working-directory: ./Exec/RegTests/EB_BackwardStepFlame/
       run: |
-        mpirun -n 2 ./PeleLMeX2d.gnu.MPI.ex eb_bfs.inp amr.max_step=2 amr.plot_int=-1 amr.check_int=-1 amrex.abort_on_unused_inputs=1 mac_proj.verbose=2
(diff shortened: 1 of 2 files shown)


baperry2:
@ThomasHowarth and @JustinFreiberger: any idea why the performance optimizations for diffusion would be causing these failures in the PeleLMeX tests?

ThomasHowarth:
The UnitTest failure doesn't look too concerning (relative error differences on the order of 10^-16). The others are weird, and I can't reproduce them on my local machine. I've had similar things happen when doing something else with the EBBFS test, where it had an MLMG failure that I couldn't reproduce locally. Have you managed to reproduce any of these errors?

d-montgomery:
I was able to reproduce the Spray test error locally, but not the MLMG errors from EBBFS.

ThomasHowarth:
Looks like a slight deviation in the multigrid solve caused the EBBFS test (which may also fix my other ongoing PR), and the missing initialisation (sorry about that) caused the spray issue. Do we reckon there are any other issues?

baperry2:
I slightly loosened the tolerances for the failing linear solve. The EBBFS case initialization has a discontinuity near an EB corner so it makes sense that it would be especially sensitive. In any event it seems to run ok after getting past the failing linear solve on the 1st timestep.

I don't think there are any more issues.

ThomasHowarth (approved):

PR #550, opened 2025-08-12
Remove using namespace amrex (1)

jrood-nrel:
I'm not liking all the inconsistency so trying to fix all of it.

### DIFF
--- a/Source/PeleLMeX.cpp
+++ b/Source/PeleLMeX.cpp
@@ -28,7 +28,7 @@ PeleLM::~PeleLM()
   typical_values.clear();
   freeProbParm();
   delete prob_parm;
-  The_Arena()->free(prob_parm_d);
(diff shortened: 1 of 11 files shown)


baperry2:
The convention that is nominally followed is `using namepace amrex` is allowed in `.cpp` files to reduce verbosity from all the `amrex::` on the variable types, but not in `.H` files, where that may cause unexpected namespace collisions in files that include that `.H`. This is similar to ERF and some other AMReX codes I believe. 

I'm open to either way. Using the namespace where possible does make the code less verbose and more readable. But if we do it in cpp but not H, it makes it annoying when moving code back and forth between them.

jrood-nrel:
I definitely think we should remove `using namespace amrex`. It makes the code inconsistent. It also makes it less readable in my opinion since you have to guess where things are coming from.

baperry2 (approved):

PR #554, opened 2025-08-14
Remove more using namespace

jrood-nrel:


baperry2 (approved):

baperry2 (changes requested):
Actually @shashankNREL is running into some issues, I think adding `amrex::` was missed for some variables.

baperry2 (approved):
I added the missing "amrex::" this should be good to go now

PR #558, opened 2025-08-23
Allow LES to use mixture averaged transport for molecular diffusion

baperry2:
Currently when LES is activated, fixed_Le and fixed_Pr are also automatically activated, so we use Prandtl, Schmidt, and Lewis numbers for both turbulent and molecular transport. Instead, we should allow fixed_Le and fixed_Pr to be off when using LES. For example, in H2 combustion, we may want nonunity Le molecular diffusion while still applying an LES model.


Source/PeleLMeX_Setup.cpp
```diff
@@ -492,34 +492,27 @@ PeleLM::readParameters()
...
+  }
+  if (m_do_les) {
     amrex::Real Schmidt = 0.7;
     pp.query("Schmidt", Schmidt);
     m_Schmidt_inv = 1.0 / Schmidt;
-  }
-  if (m_fixed_Le != 0 && !m_do_les) { // Only ask for Lewis number when not
-                                      // LES, determined by Prandtl and
-                                      // Schmidt outside of this
+    m_Lewis_inv = m_Prandtl_inv / m_Schmidt_inv;
```
d-montgomery: Should this be `m_Lewis_inv = m_Schmidt_inv / m_Prandtl_inv` so that $1/Le = Pr / Sc$ ?

baperry2: Good catch, I think you're right. This was carried over from before, but the line is also inconsistent with how `m_Schmidt_inv` is defined on line 513 below, so I've changed it.

d-montgomery (approved):

PR #561, opened 2025-09-02
SingleDropEvap test for GCM implementation

d-montgomery:
These are the scripts used to run the GCM vs. PeleMP SingleDropEvap tests shown in [PelePhysics #615](https://github.com/AMReX-Combustion/PelePhysics/pull/615)

- Compile with `SPRAY_GCM=TRUE` or `FALSE` (for PeleMP implementation)
- Run test case using the `Validate.py` script with `LiqPropsType="gcm"` or `"mp"`
- If you've run a test case with both `LiqPropsType` you can compare them using `CompareLiqPropsType.py`

This is still a work in progress and needs to be cleaned up a bit before review.


Exec/RegTests/SingleDropEvap/GNUmakefile
```diff
@@ -28,12 +28,13 @@ THREAD_SANITIZER = FALSE
...
-SPRAY_FUEL_NUM = 3
+SPRAY_FUEL_NUM = 2
+SPRAY_GCM = TRUE
 
 # PelePhysics
 Chemistry_Model = liquid_fuels_nonreacting
 Eos_Model = Fuego
 Transport_Model = Simple
 
-PELE_HOME ?= ../../..
```
baperry2, line 38: Revert this

.gitmodules
```diff
@@ -4,5 +4,5 @@
 	shallow = true
 [submodule "Submodules/PelePhysics"]
 	path = Submodules/PelePhysics
-	url = https://github.com/AMReX-Combustion/PelePhysics.git
```
baperry2, line 7: revert once PelePhysics PR is merged

CMake/BuildPelePhysicsLib.cmake
```diff
@@ -173,6 +173,7 @@ function(build_pele_physics_lib pele_physics_lib_name)
        endif()
        target_compile_definitions(${pele_physics_lib_name} PUBLIC PELE_USE_SPRAY PELELM_USE_SPRAY)
        target_compile_definitions(${pele_physics_lib_name} PUBLIC SPRAY_FUEL_NUM=${PELE_PHYSICS_SPRAY_FUEL_NUM})
+       target_compile_definitions(${pele_physics_lib_name} PUBLIC PELE_SPRAY_GCM SPRAY_GCM)
```
baperry2: Should this be `PELE_PHYSICS_SPRAY_GCM` instead of `SPRAY_GCM` to match all the CMakeLists.txt?

d-montgomery: Good catch for consistency.

d-montgomery:
I think this should be good to go now.

baperry2 (approved):

PR #556, opened 2025-08-19
Kernel fusing and stream synchronizes

ThomasHowarth:
**This is a pretty extensive PR, which may take a fair amount of time to review.**
It attempts to make everything consistent and make (micro-)optimisations where possible.
I'll start with a draft as I continue working on it. Here's a preliminary list of changes:
- `std::endl;` -> `\n`(#562)
- `for` loops: `i++`->`++i` (#562)
- lots and lots (and lots) of `const` and `constexpr`(#562)
- Removal of the `PeleLM::` namespace in several places (particularly `PeleLM::TimeStamp`) (#562)
- Consistent passing by value for `int`, `Real`,`TimeStamp` (and other small types, unless they are modified), and by reference for anything in an `amrex::Vector`, or large types. (#562)
- Fusing kernels with the more modern `amrex::ParallelFor(MultiFab...` where possible. There is more to be done here, but I'm not sure if an overload/workaround exists for working on `nodaltilebox`'s. I also need a little guidance to determine exactly when I can `amrex::Gpu::streamSynchronize()` in or outside of `lev` loops.
- Global capture (i.e. `[=]`) has been replaced with explicit capturing (#562, mostly)
- Options that are not local (e.g. `m_incompressible`) are pulled outside of `ParallelFor` loops where possible. (#562)
- Using `reserve` and `emplace_back` where possible, rather than `resize` and `define`, to construct `MultiFab`s in place. (#562)
- Lots of other micro-optimisations/code simplifications. (#562)

If any additional consistent patterns emerge, I'll add them here.

PR #562, opened 2025-09-05
Consistency, types, micro-optimisations

ThomasHowarth:
This PR is a stripped down version of [PR#556](https://github.com/AMReX-Combustion/PeleLMeX/pull/556), which does mostly non-invasive changes i.e.
- `const`/`constexpr`s
- `i++`->`++i`
- `reserve` and `emplace_back`

Kernel fusing and stream synchronizes will be left to 556.


Tests/CMakeLists.txt
```diff
@@ -96,6 +96,14 @@ function(add_test_r TEST_NAME TEST_EXE_DIR)
     set_tests_properties(${TEST_NAME} PROPERTIES TIMEOUT 18000 PROCESSORS ${PELE_NP} WORKING_DIRECTORY "${CURRENT_TEST_BINARY_DIR}/" LABELS "regression" ATTACHED_FILES "${CURRENT_TEST_BINARY_DIR}/${TEST_NAME}.log")
 endfunction(add_test_r)
 
+#Standard regression test with looser tolerances
+function(add_test_rtol TEST_NAME TEST_EXE_DIR)
+    setup_test()
+    set(RUNTIME_OPTIONS "amr.max_step=10 mac_proj.atol=5e-11 mac_proj.rtol=5e-11 mac_proj.verbose=2 ${RUNTIME_OPTIONS}")
```
baperry2: Is this still necessary? 

I think the better strategy might be to loosen the tolerances directly in the input file for the affected cases rather than create this additional test type.

baperry2:
Thanks for splitting this up, makes review much easier.

baperry2 (approved):

PR #563, opened 2025-09-06
initialize GradVel for TurbVisc

baperry2:
I think there are issues for both covered FABs and covered cells in single values FABs, so this is the best solution.


drummerdoc (approved):
Seems reasonable.

PR #565, opened 2025-09-10
Bug fix in implicit capture

SreejithNREL:
Bug fix in explicit capture of eos variable


baperry2 (approved):

baperry2:
There appear to be some additional busg at least on SYCL, but merging these fixes now.

PR #569, opened 2025-09-24
Enabling the use of Hypre for one linear system at a time

itopcuoglu:
This changes the previous mode of invoking Hypre, which checked the `AMREX_USE_HYPRE` definition for both MAC projection and for nodal projection regardless of if the input file had Hypre arguments. This meant that either both systems could be solved with Hypre, or both would use MLMG, and it was not possible to use Hypre for only one of the linear systems. Separately checking if `m_hypre_namespace_mac` and `m_hypre_namespace_nodal` exist allows for one the systems to be solved with pure MLMG while the other is solved with Hypre.


Source/PeleLMeX_Projection.cpp
```diff
@@ -481,6 +481,7 @@ PeleLM::doNodalProject(
 #endif
 
 #ifdef AMREX_USE_HYPRE
+  if(!m_hypre_namespace_nodal.empty())
```
baperry2: for formatting consistency, use `{}` for all conditionals including single lines, and run `clang-format` on the file.

itopcuoglu: I made those changes and pushed the commit.

baperry2 (approved):

PR #584, opened 2025-10-29
Remove incompressible assertion for `pelelmex_derturbforcing` 

marchdf:


jrood-nrel (approved):

baperry2:
What is the rationale here? The HIT tests are the only ones we have that run in incompressible mode, so it would be better to leave the test in that mode so it continues to be tested (also, it looks like the test is set up to require it anyway).

marchdf:
Basically the dashboard sees this : https://my.cdash.org/tests/313089995 and I am trying to fix...

baperry2:
I think it's a mistake in that ASSERT in the derive function. 
https://github.com/AMReX-Combustion/PeleLMeX/blob/1c6ef47da88d7b58b5f76100ab33d1d386db195a/Source/PeleLMeX_DeriveFunc.cpp#L1568

I think it should assert that it is incompressible to derive the forcing, not that it isn't incompressible. At least that's what would be consistent with here:
https://github.com/AMReX-Combustion/PeleLMeX/blob/1c6ef47da88d7b58b5f76100ab33d1d386db195a/Exec/RegTests/HIT/pelelmex_prob.H#L109

Probably not caught because it's an ASSERT in the derive function instead of  ALWAYS_ASSERT.

baperry2:
Or maybe that ASSERT could just be removed altogether, looking at the contents of the function it seems to be able to support either incompressible or low Mach.

marchdf:
I like that even better

marchdf:
I removed that assertion and reverted the HIT input file.

baperry2 (approved):

PR #591, opened 2025-11-13
Group species for temporals

SreejithNREL:
At times, we may need to monitor the combined mass flow rates of a set of species through temporals. For example, we might be interested to monitor the mass flow rate of air through a boundary patch 'inlet'. Currently, the code outputs the MFRs of O2 and N2 separately when the user specifies O2 and N2 in the species list of the bpatch functionality as,

 bpatch.inlet.species= O2 N2
 
 This PR lets the user specify a mixture and group species in it in the following way:
 
bpatch.inlet.species= {Air:O2,N2}

The temporal then outputs the total air MFR under the column 'inlet_Air'.


baperry2 (approved):
Looks ok to me but Copilot has a lot of suggestions on how to add error checking for the string parsing. Look through those and implement any that seem to have a good value/effort ratio. Let me know when you're done and I'll merge.

PR #600, opened 2025-12-06
revert to plotting viscturb from old but copy new2old on init

### LINKED ISSUE #599: visc_turb_fc never updated "at new time level" resulting in wrong output in plot files
Hello,

calcTurbViscosity is called with AmrOldTime only. While this is fine for the solver, the plot function uses m_leveldata_new[lev]->visc_turb_fc[...].const_arrays(). It seems that the up-to-date turbulent viscosity is therefore never outputted in the plot files. I add the copy of visc_turb_fc in copyTransportOldToNew to fix the issue, not sure it's the best solution since we only need this when we plot and not at every iteration. Do you have any better solution?

Thanks

baperry2:
Closes #599. @QuentinMale let me know if this looks good to you.

### DIFF
--- a/Source/PeleLMeX.H
+++ b/Source/PeleLMeX.H
@@ -1562,6 +1562,7 @@ public:
 
   bool checkForNaNs();
   void copyTransportOldToNew();
+  void copyTurbTransportNewToOld();
   void copyStateNewToOld(const int nGhost = 0);
   void copyPressNewToOld();
   void copyStateOldToNew(const int nGhost = 0);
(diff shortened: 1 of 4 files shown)


QuentinMale:
@baperry2 Hello Bruce,
With this solution, the velocity field and the turbulent viscosity field stored in the instantaneous solution do not correspond to the same iteration, do they? The velocity field is at AmrNewTime and the turbulent viscosity field is calculated from the strain tensor at AmrOldTime. Why not call calcTurbViscosity(AmrNewTime) at the beginning of WritePlotFile()? This way everything is at the same iteration in the solution plot file.

baperry2:
Yeah, in this solution the viscturb and velocity field will be at different timesteps in the plotfile. This may be desirable in some circumstances for debugging, but is overall confusing/counterintuitive. Your solution is better @QuentinMale.

PR #603, opened 2025-12-08
correct fake_state in initActiveControl()

QuentinMale:
impossible to run in debug mode due to the incomplete declaration of fake_state in initActiveControl(), propose a solution to be able to run in debug mode with a proper declaration of fake_state variable


baperry2 (approved):
This is a bit sketchy because it gives the Array4 valid bounds when the underlying data is still null, but it's needed given the limitations of the CellData class, and we never actually access the underlying data anyway because the s_in parameter isn't used in the relevant bcnormal for TripleFlame where this gets used.

PR #606, opened 2025-12-15
Option to specify path for temporals

d-montgomery:
This PR makes it so users can specify the name/path for the temporals diagnostics. This is especially helpful when doing parameter sweeps using the same case setup, but with different input file parameters.


baperry2 (approved):
Nice idea!

d-montgomery:
> Nice idea!

Thanks!

PR #607, opened 2025-12-16
Fix how the species fluxes are summed

d-montgomery:
This PR fixes how the species fluxes are summed for each group in speciesBalancePatch(). 

The old code was using `j = 0, 1, 2...` directly to index speciesFlux, but it needed to:
1. Get the mechanism species index from `speciesinGroup[i]`
2. Find which local index `k`
3. Use that `k` to get the correct flux from `speciesFlux[k]`

This fix ensures that fluxes are correctly summed for all group definitions, including multiple groups like {Air:O2,N2} {Fuel:POSF10264}


baperry2 (approved):
Looks good to me but @SreejithNREL please confirm that this is consistent with what you intended for the bpatches before we merge.

SreejithNREL (approved):

PR #609, opened 2025-12-31
fix species flux summing so final Dnp1 isnt double counted

baperry2:
In `PeleLM::computeDifferentialDiffusionTerms` we [do a check](https://github.com/AMReX-Combustion/PeleLMeX/blob/019098b933d1ae5ba7c99c75e9ff3ed29a67582a/Source/PeleLMeX_Diffusion.cpp#L132) on whether we are at the 0th or final SDC iteration to determine whether the species fluxes computed should be added for flux tracking purposes ($D^n$ and $D^{n+1,k}$) in the species diffusion update below).

Screenshot 2025-12-30 at 4 59 57 PM

However, `PeleLM::computeDifferentialDiffusionTerms` also gets called after the SDC iteration loop while computing the final divu for the velocity update. Currently, this trips the check on being at the final SDC iteration, and the flux is spuriously added into the flux trackers for a second time. This PR fixes the issue by changing the SDC iteration loop to directly update `m_sdcIter`, so once the SDC iteration loop completes this value exceeds m_nSDCmax and the check does not trip.

Hat tip to @wengzf20 for catching this in #608. @wengzf20 please confirm this change addresses the issue you are seeing.

### DIFF
--- a/Source/PeleLMeX.H
+++ b/Source/PeleLMeX.H
@@ -609,7 +609,8 @@ public:
   void computeDifferentialDiffusionTerms(
     const PeleLM::TimeStamp a_time,
     const std::unique_ptr<AdvanceDiffData>& diffData,
-    const int is_init = 0);
+    const int is_init = 0,
+    const amrex::Real flux_tracking_factor = 0.0);
 
   /**
    * \brief Add the Wbar contribution to the face-centered species diffusion
--- a/Source/PeleLMeX_Advance.cpp
+++ b/Source/PeleLMeX_Advance.cpp
@@ -129,7 +129,8 @@ PeleLM::Advance(const int is_initIter)
     BL_PROFILE_VAR_STOP(PLM_MAC);
     //----------------------------------------------------------------
     BL_PROFILE_VAR("PeleLMeX::advance::diffusion", PLM_DIFF);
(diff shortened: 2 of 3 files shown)


drummerdoc (commented):
This seems to avoid the observed issue, but I'm concerned about the level of "corporate knowledge" needed to understand what's happening here.  Would it be easier to understand if there was a variable `a_sdc_diffTerm_weight` that was passed in by caller so that the intention is stated clearly?

wengzf20:
Using `m_sdcIter` and disabling `m_sdcIter = sdcIter` do avoid revising rhoYFluxes during the `calcDivU()` call. The results look much more satisfactory in my case. Thank you @baperry2

baperry2:
@drummerdoc I implemented your suggestion.

drummerdoc (approved):

drummerdoc:
Nice.  Thanks!

baperry2:
@d-montgomery and @SreejithNREL: this may slightly impact the mass flow rate calculations for the SATF cases.

d-montgomery:
> @d-montgomery and @SreejithNREL: this may slightly impact the mass flow rate calculations for the SATF cases.

Thanks for the heads up @baperry2 !

PR #611, opened 2026-01-05
Assert a vector is not empty to avoid potential null pointer deference warning

jrood-nrel:


baperry2 (approved):
The null dereference should be impossible in practice anyway, but it's fine to double check that to avoid the warning.

I noticed some code here could be simplified.

PR #612, opened 2026-01-06
give option to specify custom chemistry_model paths

baperry2:
Currently, users specify a `Chemistry_Model` from those found in `PelePhysics/Mechanism`. This PR gives the user the option to activate `USE_CUSTOM_CHEMISTRY = TRUE`, in which case `Chemistry_Model` is interpreted as a path to an arbitrary location. The goal is to make it easier for users to use nonstandard chemistry models more easily without having to make modifications to the PelePhysics submodule.

### DIFF
--- a/Docs/sphinx/manual/Tutorials_BFSFlame.rst
+++ b/Docs/sphinx/manual/Tutorials_BFSFlame.rst
@@ -226,7 +226,9 @@ the relevant files, for example: ::
 
    Chemistry_Model = drm19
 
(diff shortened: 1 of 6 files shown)


d-montgomery:
I needed this 2 weeks ago! Thanks for adding this option.

d-montgomery (approved):
Works as expected and is a nice improvement to usability. My only suggestion, which doesn't need to be part of this PR, is that it would be nice if there was a list of some of the GNUmakefile options in `LMeXControls.rst` for quicker referencing.

PR #616, opened 2026-01-21
Changes to compile for single precision

jrood-nrel:


Exec/Production/ChallengeProblem/pelelmex_prob.H
```diff
@@ -69,7 +69,8 @@ time_eta(
   eta_time = 0.5 * (std::tanh((time - start) / (duration * tau)) *
                       std::tanh((-time + start + duration) / (duration * tau)) +
                     1.0);
-  eta_time = std::min(std::max(eta_time, 0.0), 1.0);
+  eta_time =
```
baperry2: Replace with `amrex::Clamp` for better readability. (same with other instances below where sequential min/max are used instead of Clamp)

baperry2 (approved):
We will need to change the names of the required tests for PRs to merge.

PR #617, opened 2026-01-21
add capability option for single output prefix across all output types

baperry2:
New usability feature: enable a common prefix across all output types (checkpoints, plot files, diagnostics, and temporals) with the `peleLM.base_output_prefix` option. As with standard output specification, this prefix may be a path (end in/contain `/`) which will be created if needed. To avoid changes in PelePhysics, this prefix is not used for diagnostics where the `.file` input option is used to give the diagnostic a custom output location.

### DIFF
--- a/Docs/sphinx/manual/LMeXControls.rst
+++ b/Docs/sphinx/manual/LMeXControls.rst
@@ -112,6 +112,7 @@ IO parameters
 ::
 
     #--------------------------IO CONTROL--------------------------
+    peleLM.base_output_prefix = outdir/    # [OPT, DEF=""] Prefix applied across all output types (plt, chk, diagnostics, temporals)
     amr.plot_int         = 20              # [OPT, DEF=-1] Frequency (as step #) for writing plot file
     amr.plot_overwrite   = false           # [OPT, DEF=false] Overwrite plot files with same name if present
     amr.plot_init_state  = false           # [OPT, DEF=false] Create a plot file during initialization before the initial projections
--- a/Source/PeleLMeX.H
+++ b/Source/PeleLMeX.H
@@ -1953,7 +1953,7 @@ public:
   std::string pltfileSource{"LM"};
   std::string m_plot_file{"plt"};
   std::string m_check_file{"chk"};
-  std::string m_base_output_pref{""};
+  std::string m_base_output_pref;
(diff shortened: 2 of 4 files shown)


d-montgomery:
Thanks for adding this capability. It works as expected and helps users who are running cases with multiple parameter sweeps. I like the overall idea, but I think it should be a directory `peleLM.base_output_dir` or  `peleLM.output_dir` that all output is directed to, including diagnostics, instead of a prefix. My primary argument is that ParaView will not open plot files that don't start specifically with `plt*`, so users could get into trouble with this feature by creating plot files that all need to be renamed. By forcing `peleLM.output_dir` to contain `/`, it would work with diagnostics too.

baperry2:
I initially coded this up as `base_output_dir`, but decided to change it to prefix to be more general - if users want the prefix to be a directory, they can just have it end with `/`. It does expose the ability for users to get into an annoying situation by having plt files be named in an inconvenient way, but users are already can do that by changing the plt file name directly, and I think Visit has slightly different expectations for plt file naming anyway.

The feature does work for diagnostics by default as is. It only gets overridden when users choose to specify the optional diagnostic file location (`.file`) which is mostly unnecessary because the diagnostics can be named whatever you want anyway. To modify this behavior, I would need to change the interface to PelePhysics and the way Diagnostic output is set up within PelePhysics.

d-montgomery:
> I initially coded this up as `base_output_dir`, but decided to change it to prefix to be more general - if users want the prefix to be a directory, they can just have it end with `/`. It does expose the ability for users to get into an annoying situation by having plt files be named in an inconvenient way, but users are already can do that by changing the plt file name directly, and I think Visit has slightly different expectations for plt file naming anyway.
> 
> The feature does work for diagnostics by default as is. It only gets overridden when users choose to specify the optional diagnostic file location (`.file`) which is mostly unnecessary because the diagnostics can be named whatever you want anyway. To modify this behavior, I would need to change the interface to PelePhysics and the way Diagnostic output is set up within PelePhysics.

That's fair. I would selfishly like an option where a base directory could be used as a prefix with `.file` in diagnostics, but didn't realize diagnostics were set up within PelePhysics.

d-montgomery (approved):

PR #624, opened 2026-02-16
Add a derived variable for element fractions

baperry2:
Only supported for Fuego and SRK EOS.

Will need https://github.com/AMReX-Combustion/PelePhysics/pull/653


d-montgomery (approved):
Thanks for adding this!

PR #625, opened 2026-02-18
decrease verbosity level needed to display regrid stats

baperry2:
Display grid sizes when regridding for verbosity 2 instead of verbosity 3.

Display total number of cells at each timestep.


d-montgomery (approved):

d-montgomery (approved):
Thanks for adding this count!

PR #627, opened 2026-02-21
deltaT iters: recompute RhoH before starting

baperry2:
In the diffusion update, we first implicitly update species diffusion, then update temperature diffusion using a linearly implicit iterative approach where the update is done in terms of temperature rather than enthalpy, and iterated until the enthalpy of the updated system converges. In the past, we've observed that the residual in this iterative solve increases on the first step before decreasing and converging (usually).

The reason for this odd behavior is that the mixture enthalpy of the new state is not updated after species are diffused and updated to the new state. Because of the chemical enthalpy of species, this diffusion affects the mixture enthalpy. If we neglect it, we start out with an inconsistent enthalpy state for the deltaT iteration process, which gets corrected when enthalpy is recomputed after the first iteration (leading to the jump is residual on the second iteration). By updating the mixture enthalpy before the iteration process starts, we start out with a consistent value, and the iterations converge monotonically. Enthalpy is updated after each iteration anyway and this does not affect the value that the enthalpy converges to.

As a result of these modifications, deltaT iterations converge typically in one less iteration, avoiding one MLMG solve, which has a mild performance benefit.

It is not yet known whether this adjustment will solve the occasional issue where convergence stalls with the LiDryer mechanism, but I expect that is a separate issue and will not be affected.

Demonstration of improved convergence for the FlameSheet case with DRM19 run for 200 timesteps with 2 SDC iterations:
FlameSheet-DRM19-SDCiter 1
FlameSheet-DRM19-SDCiter 2

Demonstration for a jet in crossflow H2 flame with LiDryer run for 200 timesteps with 1 SDC iteration:
JICFflame-LiDryer-SDCiter 1

### DIFF
--- a/Source/PeleLMeX_Diffusion.cpp
+++ b/Source/PeleLMeX_Diffusion.cpp
@@ -1,3 +1,4 @@
+
 #include <PeleLMeX.H>
 #include <PeleLMeX_Utils.H>
 #include <PeleLMeX_K.H>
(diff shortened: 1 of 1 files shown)


esclapez:
Nice ! Good catch Bruce.

d-montgomery (approved):
I'll echo @esclapez, good catch, @baperry2!

drummerdoc:
Ahh, that's excellent!  Still might need to add another step to incorporate the time-explicit parts of the update in the enthalpy equation, but this is great!

Issue #633, opened 2026-03-03
Broken triple flame ?


baperry2:
Thanks for pointing this out @EnnaDelfen! I can reproduce and see the same commit as the culprit. The velocity update/nodal projection seems to be the first thing to go haywire: the initial iterations are ~the same, the plt00000 file has small diffs in reaction rate but nothing too concerning, then the plt00001 file shows diffs in velocity and gradp, then everything starts showing significant diffs beyond that. 

Investigating further...

### Discussed in https://github.com/AMReX-Combustion/PeleLMeX/discussions/632

<div type='discussions-op-text'>

<sup>Originally posted by **EnnaDelfen** March  2, 2026</sup>
Hi all, 

In the latest version (Feb 26 41e51c1ff7f8), the triple flame seems to be broken. I am investigating but in case anyone from the current development team has a clue as to what the issue may be it may be quicker ?
To be more precise, the dtConv is stuck to 7.731042256e-08, and the solution exhibits weird kinks on the velocity profile at coarse fine boundaries

EDIT: it seems the regtest stopped working after this massive commit c8b303097156e18 </div>

baperry2:
After figuring this out, I will update the PeleLMeX CI to check for diffs on all RegTests, not just the UnitTests, as clearly that was not enough to catch this issue. (Similar to https://github.com/AMReX-Combustion/PeleC/pull/917)

baperry2:
This was fixed in #634 and #635 should prevent recurrences of similar problems.

PR #634, opened 2026-03-03
Velocity Projection Bug Fix

baperry2:
Proposed fix for #633. Looks like old and new states got swapped during #556.


EnnaDelfen (approved):
This seems to have solved the issue on my end and the log file for my runs are now similar to what I had in my previous November version (incl. flamesheet TC). Can't really comment on the appropriateness of the format of the coding but it seems fairly straightforward. Thanks for attending to this so quickly.

ThomasHowarth (approved):
Thanks for catching this Bruce - not sure how this got lost in translation in the previous PR; apologies.

baperry2:
I'll release another minor version soon because this is a substantial change.

PR #635, opened 2026-03-04
run fcompare in CI

baperry2:
Build a reference version of PeleLMeX and compare results for RegTests during CI. Mirrors https://github.com/AMReX-Combustion/PeleC/pull/917 and what is done for the UnitTests.


baperry2:
Confirmed that this PR does catch intentional bugs (which have now been reverted): https://github.com/AMReX-Combustion/PeleLMeX/actions/runs/22740480520/job/65952674740

unitTest should still be failing because I added additional output variables that will not be in the output of the reference version (but all existing variables should now match with the intentional bugs reverted).

jrood-nrel (approved):

PR #636, opened 2026-03-06
Minor fix so the postAdvance function works on GPU

d-montgomery:
This updates how prob_parm is passed to the `postAdvance` function in the ProblemSpecificFunctions for compatibility with CPU and  GPU.

### DIFF
--- a/Docs/sphinx/manual/Troubleshooting.rst
+++ b/Docs/sphinx/manual/Troubleshooting.rst
@@ -208,7 +208,7 @@ initial solution is most likely erroneous.
 
 In the case of the second message, the system was integrated up to 2.459e-6 s, but CVODE was not able
 to proceed any further as its internal step size dropped to a small value. This could indicates that your
-CFL condition is too loose and the chemical stifness can't be properly handled by
+CFL condition is too loose and the chemical stiffness can't be properly handled by
 CVODE. You can consider reduce your CFL number:
 
 ::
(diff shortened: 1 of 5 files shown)


baperry2:
We should definitely pass the device copy of `prob_parm` as that is likely to be needed by users. `prob_parm` is really intended to contain data used on device and the host version mostly just exists as a pass through to create the device copy. 

We should either pass both host and device copies, which makes the API a bit ugly, or pass only the device copy, which may force users to do some operations in device code that would be more efficient on host. For example, if a binary flag indicates whether or not an operation should be carried out, it would be most efficient to do that check once and never launch a ParallelFor if the operation is not to be carried out, rather than launching the parallel for and doing an identical check at each grid point.

baperry2:
Leaning toward passing both host and device copies as the better option. It's a bit off label, but users may want to stick something in `prob_parm` to determine functionality that doesn't directly involve device code (e.g., a flag indicating whether or not to dump extra output files)

d-montgomery:
> Leaning toward passing both host and device copies as the better option. It's a bit off label, but users may want to stick something in `prob_parm` to determine functionality that doesn't directly involve device code (e.g., a flag indicating whether or not to dump extra output files)

This is exactly the issue I'm running into. I need `prob_parm` as a flag prior to doing any device code.

baperry2 (approved):
Thanks for including clear comments documenting the API

d-montgomery:
It's just as much for me as a reminder!

d-montgomery:
I snuck in a minor change to the README that reflects @jrood-nrel recent update from c++17 to c++20.

PR #644, opened 2026-03-25
Deallocate in BPatch destructor

jrood-nrel:


baperry2 (approved):
It does look like this was memory leaking due to not actually deallocating the boundary patch data.

I also removed some unused pointers related to bpatches and updates the test that is supposed to include this capability (the wrong input file had the inputs).

PR #645, opened 2026-03-25
Guard against divide by 0 in SwirlFlowWallInteractions

jrood-nrel:


baperry2 (approved):
Looks fine. This case is not really used to my knowledge (and isn't tested), so this is not a big deal, but might as well add a check.

PR #646, opened 2026-03-25
fix variable swap in divu dt estimate

### LINKED ISSUE #642: Parameters swapped in Timestep.cpp
https://github.com/AMReX-Combustion/PeleLMeX/blob/cd70011ad60b8103a9cf3c44ac46efcf137dce10/Source/PeleLMeX_Timestep.cpp#L200-L205

It appears `divu` and `vel` are swapped here.

baperry2:
Closes #642

We don't have any tests that use the divu_dt method in question, but if used in debug mode it currently results in an OOB error. 

This PR fixes the variable swap and sets a test to use the method.


jrood-nrel (approved):

baperry2:
The modifications here change the initial timestep size in the affected test, so the slight diffs are expected. @jrood-nrel the gold files for the nightly for the test will need to be updated.

PR #670, opened 2026-06-09
BUGFIX: set Dirichlet values in grow cells prior to setLevelBC

drummerdoc:
BUGFIX: set Dirichlet values in grow cells prior to setLevelBC so that inhomogeneous values are properly loaded

Identified while working on a Lid-driven cavity test


jrood-nrel (approved):

d-montgomery:
Thanks for finding this, @drummerdoc! Would you mind just cleaning up the comments a bit since they reference a RegTest that doesn't yet exist in the development branch?

drummerdoc:
Oops!  Dang. Good catch. I'll add that test without mesh mapping to keep things orthogonal.

Issue #674, opened 2026-06-26
Recompute TEMP/RHORT from conserved state after average-down (all cases)


drummerdoc:
**Context**. The mesh-mapping AMR work made averageDownState/averageDownScalars recompute TEMP and RHORT via the EOS (setTemperature/setThermoPress) immediately after the conservative restriction of the conserved densities (rho, rhoY, rhoH), so the derived thermo state is consistent with the restricted conserved set. This is currently gated on m_mesh_mapping; the default (non-mapped) path still arithmetic-averages TEMP at the average-down and only recomputes at the scheduled EOS point later in the step.
**Proposal**. Apply the post-restriction EOS recompute on the non-mapped path as well — it's more thermodynamically consistent (T derived from the restricted rhoY/rhoH rather than linearly averaged).
**Risks / why it's not in the mesh-mapping PR.**

Changes coarse-cell TEMP/RHORT for existing reacting-AMR runs → breaks fcompare golds and may perturb user-tuned cases. Needs gold regeneration + a heads-up.
Adds an EOS Newton solve at each average-down; should likely be restricted to the covered coarse region (needs a covered-cell mask) before enabling broadly.

**Where**. Source/PeleLMeX.cpp — averageDownState, averageDownScalars.

PR #656, opened 2026-04-29
Mesh mapping

drummerdoc:
Implement single-level stretched orthogonal grid-aligned meshes using a strategy that follows the one used in AmrWind. This is a rather massive PR. To keep the tasks manageable, the work was committed incrementally with lots of checking at each step. It was co-authored with Claude. A follow-on set of runs is currently underway to evaluate the performance of this feature to efficiently capture wall-bounded turbulence in a Cartesian channel. If successful, later improvements will extend the implementation to AMR (max_level>0) and allow interaction with the various tools to generating and inflowing turbulent fluctuations.  See the documentation of the sub-commits that explain each of the steps taken and the tests used to confirm correctness.

Copilot should probably make the first pass for PR review before a human is subjected to this mess.

PR #675, opened 2026-06-26
Generalize mesh mapping to incorporate AMR

drummerdoc:
**Summary**
Extends the mesh-mapping capability from #656 to work correctly with AMR. Under a coordinate-aligned map, coarse/fine transfer, the projection sync, and the conservation diagnostics must account for the cell-volume metric (`detJ`) and per-direction stretch (`fac`). This PR implements that, hardens the regrid/interpolator paths, adds incompressible and reacting regression tests, and documents usage.

**Dependency**
Requires AMReX #5500 (MLNodeLaplacian: aniso-sigma across coarse/fine boundaries), needed for the anisotropic σ that mesh mapping produces at C/F boundaries. Pulled in by bumping `Submodules/PelePhysics` → `4abf909` (PelePhysics #681), which pins amrex `6a6dcd9`. PeleLMeX has no direct amrex pin, so a recursive checkout resolves it transitively.

**What's included**
*Conservative coarse/fine transfer*
- `detJ`-weighted (volume-conservative) restriction of the conserved densities (`rho`, `rhoY`, `rhoH`); Xi-space restriction for velocity.
- `MeshMappedCellConsInterp` generalized from velocity-only to the full state via per-component Xi-space weights (`detJ`/`fac` for velocity, `detJ` for conserved densities, identity otherwise) — one interpolator serves incompressible and reacting state. Fine ghost cells are now un-weighted (previously left in Xi-space).
- `TEMP`/`RHORT` recomputed from the restricted conserved state via the EOS (mesh-mapping path only).

*Robustness*
`getInterpolator()` now takes the coarse level explicitly, removing the mutable side-channel.
During regrid, mesh-map metrics and interpolators are built *before* the coarse→fine fill, and sized from the MeshMap rather than the stale `finest_level`.

*Diagnostics*
Mass/species/energy balance integrals are detJ-weighted, so `d/dt(∫field) == net boundary flux` holds under mapping.

*Tests*
Incompressible: `LidDrivenCavity` mapped-AMR (TanhStretch + 1 AMR level), via a dedicated CI job that builds Hypre from source.
Reacting: `FlameSheet` mesh-mapping+AMR conservation regression (lateral TanhStretch, AMR on the perturbed flame), with an identity-map control and a flux-balance `test.py`.

*Docs*
Manual (`Model.rst`): the `initdata_mapped` signature, a warning that any coordinate-dependent code (including external analysis) must transform index→physical and transform vectors/operators, and the metric/operator transform relations.

**Known limitations / future work**
Conservation is exercised by the new tests but not yet by a committed CI reference comparison.
The EOS recompute after restriction is enabled only on the mesh-mapping path; generalizing it to all cases is worthwhile but result-shifting — tracked in #674.
That recompute currently spans the whole domain; restricting it to the covered coarse region is a noted perf follow-up.
Mesh mapping remains incompatible with EB, RZ, plasma, and turbulent inflow (unchanged from #656).

**Testing**
- Identity-map control conserves to ~4e-9 per-step / round-off cumulative; the stretched case meets the same tolerance.
- New `CPU-Hypre-MeshMapping` CI job; FlameSheet conservation tests run in the standard CMake matrix.


.github/workflows/ci.yml
```diff
@@ -135,16 +135,16 @@ jobs:
...
-          fi
       - name: Dependencies
         run: |
           # Install MPI
           ${{matrix.install_deps}}
           # Install Python packages
-          python -m pip install --upgrade pip
-          pip install pynose numpy pandas
+          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install pytest numpy pandas
```
d-montgomery: Try adding pynose back here. I think the tests are failing because Cmake from the development branch calls nosetest.

drummerdoc: Thanks for looking, but in my local version of [10785a2] there are no instances of "nosetests" that I can find, as I replaced them with `python -m pytest` as suggested.  If you look at the last couple of commits to that branch, it put me a cycle of removing and changing the tester...now it seems confused.

d-montgomery (approved):

PR #677, opened 2026-07-01
Fix turb conv distance

drummerdoc:
March through the turbulent-inflow data at the *time-integrated* controlled inlet velocity (∫ V_in dt) instead of `turb_conv_vel * time`, so the turbulence marcher stays consistent with the controller-driven mean inflow and introduces no position jumps when `ctrl_V_in` is adjusted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,9 @@ jobs:
             reference_comparison: ON
           - os: ubuntu-24.04
             #Note mpich on ubuntu-24.04 does not run mpiexec correctly so we are using openmpi
-            install_deps: sudo apt-get install -y libopenmpi-dev openmpi-bin
+            install_deps:
+              sudo apt-get update
+              sudo apt-get install -y libopenmpi-dev openmpi-bin
             comp: gnu
             procs: $(nproc)
             ccache_cache: ~/.cache/ccache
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -138,11 +140,18 @@ jobs:
           if [ "${RUNNER_OS}" != "macOS" ]; then
             ${{github.workspace}}/Submodules/PelePhysics/Submodules/amrex/.github/workflows/dependencies/ubuntu_free_disk_space.sh
           fi
-      - name: Dependencies
```
d-montgomery, line 141: We had to overcome this same issue in #675. I think what we did [there](https://github.com/AMReX-Combustion/PeleLMeX/pull/675/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR143) is cleaner and limits duplicate code:
~~~
- name: Dependencies
        run: |
          if [ "${RUNNER_OS}" != "macOS" ]; then
            sudo apt-get update
          else
            brew update
          fi
~~~

d-montgomery (changes requested):
Thanks for adding this @drummerdoc. I have one minor request related to the CI changes.

d-montgomery (approved):

PR #685, opened 2026-08-09
Fix recycling-plane inflow bookkeeping, slab levels, and post-regrid mean

drummerdoc:
Follow-up to #655, fixing three bookkeeping defects around the multilevel snapshot machinery:

1. n_samples was never incremented. The warmup gate could therefore never open (no fluctuations were ever injected), and in the default cumulative-average mode alpha was stuck at 1, making mean_src the previous snapshot and fluct_src a one-step temporal difference. Now seeded at 1, weighted by 1/(n_samples+1), and incremented each snapshot.
2. Levels whose grids miss the plane got no slab storage (the early-out ran before the complementIn), so a level owning the inlet but not the sampling plane injected zero fluctuation; the seed/update loops also dereferenced the null slab pointers. For lev > 0 the slab is now built entirely from complementIn boxes and filled from the coarser level.
3. Regrid-created levels seeded the running mean with zero while initialized stayed true, so their first fluctuation was the full velocity (~doubled mean inflow until the mean recovered). The mean is now seeded by PC interpolation of the coarser level’s mean, via a new shared helper interpRecyclingSlabFromCoarse.

Note that this PR changes test results for the EB_BFS cases. I have confirmed that the results now obtained in those cases are superior to the initial results (thus, this PR should be merged in spite of not passing the CI tests).

Testing: Confirmed on a half-resolution 3D BFS case (128×32×32, 16000 steps ≈ 30 flow-throughs, including a checkpoint restart at step 8000): recycled fluctuations sustain a statistically stationary turbulent state from 8.5 τ onward (KE CoV 37%, enstrophy CoV 29%, no secular trend), where the pre-fix code injects nothing and the flow remains laminar. Note: a restart mid-run on this test case exercised the checkpoint mean/n_samples path on a real problem.

### DIFF
--- a/Exec/RegTests/EB_BFS_Recycling/analyze_stationarity.py
+++ b/Exec/RegTests/EB_BFS_Recycling/analyze_stationarity.py
@@ -33,10 +33,19 @@ WINDOW_TAU = 4.0     # 4 flow-through averaging window
 def read_tempstate(path: Path):
     with path.open() as f:
         rows = list(csv.DictReader(f))
+    # A restarted run re-appends the CSV header line mid-file; skip those
+    # rows. Restarts can also duplicate iterations already in the file
+    # (e.g. restarting from a checkpoint earlier than the last row written
+    # by the previous run); keep the last occurrence of each iteration.
+    by_iter = {}
+    for r in rows:
+        if r["iter"] == "iter":
+            continue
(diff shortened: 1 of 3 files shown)


d-montgomery (approved):
Thanks @drummerdoc!

PR #686, opened 2026-08-10
Add peleLM.recycling_mode: inject full sampled velocity or fluctuations

drummerdoc:
Adds an alternative injection mode for the recycling-plane inflow:

peleLM.recycling_mode = fluctuations (default) | full

fluctuations is the existing behavior (zero-mean fluctuation about a running mean, added to the bcnormal profile). full injects the entire sampled velocity with no mean management.

Implementation reuses the existing machinery wholesale: fluct_src acts as a generic injection buffer, and in full mode updateRecyclingPlaneSnapshot copies u_src into it and skips the mean/fluctuation arithmetic. The multilevel interp cascade, copy-on-intersect, EB masking, regrid rebuild, and fillFromRecyclingPlane are unchanged. Full mode is stateless, so the slab is seeded eagerly (fresh start, restart, and a guard at the top of Advance) and valid data exists for the very first BC fill; checkpoints carry the running mean only in fluctuations mode; the warmup and averaging-window controls apply only to fluctuations mode (setup warnings if set with full).

The recycled velocity is enforced by the boundary-fill machinery itself: the mode is threaded into PeleLMCCFillExtDirState as a plain int member, and **velocity writes from bcnormal are ignored on the recycling face in full mode** — problem setups run unmodified in either mode, with no double-counting of the mean profile. Documented in LMeXControls and the BFS recycling tutorial.

Testing: default mode changes no answers, so the CI reference comparison is expected to pass.

### DIFF
--- a/Docs/sphinx/manual/LMeXControls.rst
+++ b/Docs/sphinx/manual/LMeXControls.rst
@@ -380,16 +380,32 @@ the user through the standard ``Inflow`` boundary condition; this feature only
 supplies the bracketed fluctuation. The running mean is preserved in
 checkpoints and restored across restarts.
 
+Alternatively, ``peleLM.recycling_mode = full`` injects the entire sampled
+velocity, :math:`\mathbf{u}_{\text{inlet}}(t) = \mathbf{u}_{\text{src}}(t)`,
+with no running mean maintained. The recycled velocity is enforced by the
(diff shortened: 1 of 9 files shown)


wsong1305 (approved):
Thank you for adding direct velocity mapping!

PR #688, opened 2026-08-15
Add mass-flux control to full-mode recycling

drummerdoc:
## Summary

Follow-up to #686 (`peleLM.recycling_mode = full`): add a mass-flux controller to full-mode recycling so that configurations with net heat transfer between the inlet and the sampling plane remain well-posed.

## Motivation

Full mode feeds the sampled plane velocity straight back to the inlet, so the bulk flow rate is a closed feedback loop with no memory of the intended inflow. In the low-Mach setting, any net heat exchange between the inlet and the sampling plane biases that loop: wall heat loss makes `div(u) < 0` in the duct, so the plane always returns a *smaller volume flux* than the inlet received — mass flux, not volume flux, is what convection conserves. The recycled bulk flow therefore decays monotonically, and the decay self-accelerates (slower flow → longer residence time → more cooling per pass) until the inlet reverses.

Observed concretely on a backward-facing-step configuration similar to `Exec/RegTests/EB_BFS_Recycling`, with the sampling plane just upstream of the expansion and an isothermal step top colder than the incoming gas: total inlet backflow within a few tens of flow-through times. An adiabatic constant-area duct is merely *neutrally* stable — full mode appears to work there only because the bias is zero, but nothing anchors the flow rate.

## What this does

Every snapshot, `applyRecyclingFluxControl` rescales the injection buffer by `gamma = Q0/Qs`:

- **Both flow rates are integrated over the inlet-face fluid footprint** — a 0/1 EB mask of the first interior cell layer at the inflow face, held on the level-0 slab layout and rebuilt with the slab storage. This is the cross-section the sample is actually injected into; it keeps `bcnormal` values behind an embedded boundary out of the target, and it remains correct when the sampling plane sits in a wider cross-section than the inlet (e.g. past an expansion).
- **The target `Q0` is measured once from the `bcnormal` inflow profile** (recycling preload and synthetic turbulence disabled in the measurement fill, so a `bcnormal` that increments the preloaded value measures identically), or set explicitly with `peleLM.inlet_plane_target_flow_rate`. No checkpoint state; the target is re-measured after restart.
- **The scaling is uniform across all velocity components.** The sampled fluctuation field is solenoidal to low-Mach accuracy, with continuity coupling the normal and tangential components mode by mode; uniform scaling preserves that structure (`div(gamma u) = gamma div(u)`), so the injected turbulence passes through the inlet-adjacent projection intact. A normal-only scaling would leave an irrotational residual for the projection to remove, distorting the tangential fluctuations. No-slip walls and the sample's relative turbulence intensity `u'/U` carry through unchanged.
- **Guard rails:** `gamma` is clipped to `[0.5, 2]` with a warning; a sampled through-flow below 5% of the target, or of opposite sign, aborts with an explicit message (a collapsed sample cannot be rescaled into a meaningful inflow). Opposed-inflow configurations (both faces `ext_dir` in the plane direction) disable the controller with a warning, since one bulk factor cannot serve two inlets fed from a single plane.

## Inputs

```
peleLM.inlet_plane_flux_control      = 1     # default on in full mode; 0 recovers the old behavior (with a warning)
peleLM.inlet_plane_target_flow_rate  =       # optional; default: measured from bcnormal on first use
```

Fluctuations mode is untouched (its mean is anchored by `bcnormal` by construction) and warns if the new inputs are supplied.

Documented in `LMeXControls.rst` and the BFS recycling tutorial.

### DIFF
--- a/Docs/sphinx/manual/LMeXControls.rst
+++ b/Docs/sphinx/manual/LMeXControls.rst
@@ -380,16 +380,32 @@ the user through the standard ``Inflow`` boundary condition; this feature only
 supplies the bracketed fluctuation. The running mean is preserved in
 checkpoints and restored across restarts.
 
+Alternatively, ``peleLM.recycling_mode = full`` injects the entire sampled
+velocity, :math:`\mathbf{u}_{\text{inlet}}(t) = \mathbf{u}_{\text{src}}(t)`,
+with no running mean maintained. The recycled velocity is enforced by the
+boundary-fill machinery itself: whatever ``bcnormal`` writes to the velocity
+components on the recycling face is ignored in this mode, so existing
+problem setups (including ones that add a mean profile to the velocity)
+run unmodified; ``bcnormal`` still controls the thermodynamic state. Full
+mode carries no restart state (the injection buffer reseeds from the first
+snapshot after initialization or restart, taken before the first boundary
+fill), and the ``inlet_plane_avg_window`` / ``inlet_plane_warmup_steps``
+controls are ignored.
+
 ::
 
     #-----------------------Recycling-plane inflow-----------------------
     peleLM.use_inlet_from_plane     = 0       # [OPT, DEF=0] Master switch (0 disables)
     peleLM.inlet_plane_dir          = -1      # [REQ if active] Sampling axis: 0=x, 1=y, 2=z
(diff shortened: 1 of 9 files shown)


wsong1305 (approved):
Thank you Marc for implementing the mass-flux controller. I have tested this with BFS configuration and the flow also behaves as expected!

PR #691, opened 2026-08-21
Option to coarsen base level when restarting from plt file

d-montgomery:
This PR adds support for restarting from a plt file with a 2x coarser base mesh via `peleLM.initDataPlt_coarsen`. This is useful for reducing computational cost when continuing a simulation. Currently this only works for Non-EB cases. 

This feature uses multi-level mapping so that plt file levels are shifted to match the new grid hierarchy. For example:  `plt_old` level 0 becomes the new level 1, `plt_old` level 1 becomes the new L2, etc. The base level is coarsened using conservative volume-weighted averaging and intermediate levels are created via interpolation from coarser levels.

I added an example in the HotBubble RegTest that demonstrates this capability.

Restarting HotBubble from t = 0.1 s with 2x coarser mesh
restart_coarser_level0_0 1s

Resulting meshes at t = 0.15 s after restart at t = 0.1 s
restart_coarser_level0_0 15s


drummerdoc (approved):
Thanks for this nice addition!

Tested this in 2D and 3D and after some reasonable playing inputs, the mods worked as expected.  It might be nice to pair this with something that removes the coarsest level.  Orthogonal to this add though.

Everything looks clean.

PR #698, opened 2026-09-15
bump PelePhysics to a0b0e88 

drummerdoc:
Includes amrex 5a08d94f with MLTensorOp mapping factors; DiagFramePlane fixes


drummerdoc:
The EB_ODEQty reference mismatch is round-off: all absolute differences are ≤ 2e-11; the relative errors that trip the 1e-12 gate are on near-zero fields (z_velocity ‖A‖ ≈ 1e-12, divu ≈ 1e-13, gradpz ≈ 1e-8). Density, temperature, rhoh, pressure and the ODE quantities agree to 1e-12 or better. Consistent with a change of summation order in the EB path across the 204 amrex commits in this bump (e.g. AMReX-Codes/amrex#5824, atomic adds in state redistribution). No PeleLMeX source changes in this PR.

baperry2 (approved):

drummerdoc:
The reference-comparison failures (EB_ODEQty, flamesheet-drm19-3d, and the 2D EB set on macOS) are round-off. In every table the state fields — velocity, density, temperature and the species with non-trivial mass fractions — agree with the reference to 1e-13 or better. The entries that exceed the 1e-12 relative gate are of two kinds: trace species and rates that are numerically zero (Y(CO2) differing by 3e-30 reads as a relative error of 1.5), and derived diagnostics that amplify round-off (I_R reconstructed as a state difference over dt, HeatRelease from I_R, gradp from the iterative projection at its tolerance). The 2D EB cases show more red simply because a cold flow has more near-zero fields.

This PR has no PeleLMeX source change; it moves the amrex pin by 204 commits (26.07-12 → 26.09-123), and a change of summation order somewhere in that range (e.g. atomic adds in state redistribution, AMReX-Codes/amrex#5824) is enough to produce the above. The comparison will stay red by construction until the pin is on development, since the reference build uses the old amrex. Requesting merge as a known-good dependency update.

The Lint-clang-tidy (EB-ON) job failed because clang 18 segfaults in its optimiser while compiling amrex's Src/EB/AMReX_EB_STL_utils.cpp at -O3 (new STL code from AMReX-Codes/amrex#5870). The lint job now builds as Debug (-O0), which sidesteps the optimiser without dropping EB from the lint matrix — clang-tidy works on the AST, so its findings are unchanged.

PR #703, opened 2026-09-17
Mesh/flamesheet mapped pair

drummerdoc:
Follow-up to #702. The Soret flux was the one mapped operator without a mapped-vs-unmapped round-off check: it needs a light species in a temperature gradient, which none of the mapping-aware cases had. FlameSheet has both (H, H2 in a drm19 flame) plus Simple transport, Wbar and chemistry, so a ConstantMap pair of it covers everything at once.

Commits

1. RegTests/FlameSheet: initial condition at the physical cell position under mesh mapping. initdata() placed the PMF profile and its x perturbation with the AMReX (xi-space) coordinates, so on a mapped mesh the flame sat at each cell's xi position rather than its physical one and the perturbation wavelengths scaled with the xi extent. The IC is factored into initdata_at() (physical position, physical domain lengths, physical half-cell width for the PMF cell average) and initdata_mapped(), selected automatically when a mapping is active, evaluates those from the MeshMapEvaluator. Unmapped runs are bit-identical (same expressions; with the identity map the half-cell width is taken from the grid spacing directly).

Behaviour change: flamesheet-drm19-mapped-amr-3d now initialises its flame where the physical mesh is — the intent of that input — so its reference comparison is red once on this PR.

2. Tests: FlameSheet ConstantMap pair — Soret, Wbar and mixture-averaged D under mapping. flamesheet-mapped-2d.inp (ConstantMap 2 in x) and _ref.inp (identity) are the same physical flame on index-identical grids: 32x64, 5 steps at fixed dt, use_wbar = 1, transport.use_soret = 1, CVODE. A new CMake function add_test_mapped_pair() runs both and requires the final plotfiles to agree field by field with compare_plotfiles.py (#702). Registered in 2D, non-EB, as flamesheet-mapped-2d.

3. Tests: FlameSheet mapped pair — tolerance 1e-8 for the reacting case, no I_R in the plotfiles. The tolerance is an argument of add_test_mapped_pair(): 1e-10 for a non-reacting case, 1e-8 here. CVODE's adaptive stepping amplifies ~1e-13 round-off differences in the incoming state to ~1e-9 in the reaction increment, so I_R (the increment over dt) and what derives from it (divu, HeatRelease) sit at 1e-9 while every state field agrees far tighter; I_R is left out of the plotfiles (peleLM.plot_react = 0). A mapping error in a flux would show at 1e-4 or worse after 5 steps.

Results (macOS, llvm, MPI x2)

flamesheet-mapped-2d vs _ref, 5 steps, both runs reporting "mixture-averaged transport with Soret effects": density 3e-13, rhoh 4e-13, temp 2.5e-12, major species (Y(O2), Y(H2O), Y(CH4), Y(CO2), Y(N2)) 1e-12 to 5e-12, Y(H2) 1.7e-12, Y(H) 7e-11; the largest state-field difference is 2.2e-10 on the trace species Y(HCO) (max 8.6e-6). divu 2.3e-10, gradp (after the fac scaling) 1.5e-9, avg_pressure 1.4e-10; before plot_react = 0 the I_R fields sat at 1e-10 to 4e-9. PASS at 1e-8.

4. RegTests: compare_plotfiles.py accumulates max|ref| and max|diff| again — restores the two accumulations displaced by #702's non-finite check; without them the gate passed vacuously.


drummerdoc:
Good catch on the comparator: the non-finite check that went in during #702's review displaced the two max() accumulations, so the merged script reported zero for every field. Restored in a3cf4a1 ; the numbers quoted in the description were produced with the accumulating version, and re-running it on the same plotfiles reproduces them. On the second point, 2.2e-10 is already the diff/scale column (max|diff| = 1.9e-15 against max|ref| = 8.6e-6), so it is within the 1e-8 tolerance.

